PR #12, opened 2020-05-13
new-scala-file command

landerlo:
WIP new-scala-file. Still needs some polishing but this is a working new-scala-file command.

### DIFF
--- a/lua/metals.lua
+++ b/lua/metals.lua
@@ -5,9 +5,9 @@ local util = require'metals.util'
 local M = {}
 
 local function execute_command(command, callback)
-  vim.lsp.buf_request(0, 'workspace/executeCommand', command, function(err, _, resp)
+  vim.lsp.buf_request(0, 'workspace/executeCommand', command, function(err, method, resp)
     if callback then
-      callback(err, resp)
+      callback(err, method, resp)
     elseif err then
       print('Could not execute command: ' .. err.message)
     end
@@ -60,6 +60,47 @@ end
 M.sources_scan = function()
   execute_command({
     command = 'metals.sources-scan';
+  })
+end
+
+M["metals/quickPick"] = function(_, _, resp)
+  local ids = {}
+  local labels = {}
+  for i, item in pairs(resp.items) do
+    table.insert(ids, item.id)
+    table.insert(labels, i .. '- ' .. item.label )
+  end
+
+  local choice = util.input_list(labels)
+  if (choice == 0) then
+    print("\nmetals: operation cancelled")
+    return { cancelled = true; }
+  else
+    return { itemId = ids[choice] }
+  end
+end
+
+M['metals/inputBox'] = function(_, _, resp)
+    local name = util.input_box(resp.prompt .. ': ')
+
+    if (name == '') then
+      print("\nmetals: operation cancelled")
+      return { cancelled = true; }
+    else
+      return { value = name; }
+    end
+end
+
+M['metals/executeClientCommand'] = function(_, _, cmd_request)
+  if cmd_request.command == 'metals-goto-location' then
+    lsp.util.jump_to_location(cmd_request.arguments[1])
+  end
+end
+
+M.new_scala_file = function(directory_opt)
+  execute_command({
+    command = 'metals.new-scala-file';
+    arguments = { directory = directory_opt}
   })
 end
 
--- a/lua/metals.lua
+++ b/lua/metals.lua
@@ -68,7 +68,7 @@ M["metals/quickPick"] = function(_, _, resp)
   local labels = {}
   for i, item in pairs(resp.items) do
     table.insert(ids, item.id)
-    table.insert(labels, i .. '- ' .. item.label )
+    table.insert(labels, i .. ' - ' .. item.label )
   end
 
   local choice = util.input_list(labels)
@@ -97,11 +97,27 @@ M['metals/executeClientCommand'] = function(_, _, cmd_request)
   end
 end
 
-M.new_scala_file = function(directory_opt)
+--[[
+directory_uri_opt: Path URI for the new file. Defaults to current path. e.g. 'file:///home/...'
+name_opt: Name for the scala file. e.g.: 'MyNewClass'. If nil, it's asked in an input box.
+--]]
+M.new_scala_file = function(directory_uri_opt, name_opt)
+  local args_string_array = {}
+  if directory_uri_opt then
+    table.insert(args_string_array, 1, directory_uri_opt)
+  else
+    table.insert(args_string_array, 1, vim.NIL)
+  end
+  if name_opt then
+    table.insert(args_string_array, 2, name_opt)
+  else
+    table.insert(args_string_array, 2, vim.NIL)
+  end
+
   execute_command({
-    command = 'metals.new-scala-file';
-    arguments = { directory = directory_opt}
-  })
+      command = 'metals.new-scala-file';
+      arguments = args_string_array
+    })
 end
 
 -- Thanks to @clason for this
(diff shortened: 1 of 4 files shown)


ckipp01:
Thanks a lot for starting to work on this! How does this actually work though? Looking at it, I'd expecting that after the `executeCommand` is called (`metals.new-scala-file`) then Metals will respond with a `window/showMessageRequest` with the possible options. However, right now that's unhandled.

landerlo:
Sorry, I forgot to update the setup file, I pushed the changes again with the new setup you need to change the setup as in the new setup example I pushed,  declaring the inputBox and quickPick capabilities, so metals uses that.

The cancelling also works now.

Unfortunately this metals feature should be more standard and use workspaceEdits instead of creating the file in the filesystem and editing it by itself. This means that we never get which file was created or when, so we cannot jump to the new file after creation, that must be using vscode internals to do that. Maybe we can get the trigger by the peeking the status and message messages from metals, but it's not trivial. The sane thing would have been to send workspaceEdits to a new file and let the editor do the right thing and get the focus.

landerlo:
Ok, the focus change is completed now, so this completes the functionality.
I was missing it uses GoToLocation commands to change focus.
You need to update your setup to the example setup here.

We'll need to think a way to have to manage setup and not having to change user config , when metals-nvim add endpoints or capabilities like in this PR. 

Maybe instead of having to using the lua << whole setup code, have something like setup_nvim_metals { myOverrides }

lua/metals.lua
```diff
@@ -63,6 +63,47 @@ M.sources_scan = function()
   })
 end
 
+M["metals/quickPick"] = function(_, _, resp)
```
ckipp01, line 66: Ahh good idea doing it this way to match Nvim `M["metals/quickPick"].

lua/metals.lua
```diff
@@ -63,6 +63,47 @@ M.sources_scan = function()
...
+  local ids = {}
+  local labels = {}
+  for i, item in pairs(resp.items) do
+    table.insert(ids, item.id)
+    table.insert(labels, i .. '- ' .. item.label )
```
ckipp01, line 71:
```suggestion
    table.insert(labels, i .. ' - ' .. item.label )
```

Nitpick, can we add a space here as well. I think it looks a bit better this way.

lua/metals.lua
```diff
@@ -63,6 +63,47 @@ M.sources_scan = function()
...
+    end
+end
+
+M['metals/executeClientCommand'] = function(_, _, cmd_request)
+  if cmd_request.command == 'metals-goto-location' then
+    lsp.util.jump_to_location(cmd_request.arguments[1])
+  end
+end
+
+M.new_scala_file = function(directory_opt)
```
ckipp01, line 100: In what scenario is the `directory_opt` used here? Is this mainly for someone to use if they want to pass in a parameter?

lua/metals/util.lua
```diff
@@ -120,4 +120,30 @@ M.path = (function()
...
+  return vim.fn.input(prompt)
+end
+
+M.input_list = function( options)
+  return vim.fn.inputlist(options)
+end
+---------------------------------------------------------------------
+
+-- TODO: See if nvim provides similar debugging function
+M.to_string = function(o)
```
ckipp01, line 135: More of a general question due to my lack of lua knowledge, but how are functions like this normally handled? I see this want like a `private[metals]`, but I'm assuming there isn't one? So anyone using the plugin can reach into `util` and use this if they want right?

ckipp01:
> I removed my 'rant' about metals not using workspaceEdits to create the file (I still think it'd have been better design than creating the file outside).

I don't exactly remember the reason, but I remember we had this conversation on the Metals side of things I believe and obviously didn't go that route. I don't remember the exact reasoning.

> We'll need to think a way to have to manage setup and not having to change user config , when metals-nvim add endpoints or capabilities like in this PR.

> Maybe instead of having to using the lua << whole setup code, have something like setup_nvim_metals { myOverrides } 

Yea I've been thinking about this a bit lately and trying to figure out what a good balance between just offering everything and having the user set it up, or just doing it for them. I've noticed some plugins like `completion-nvim` for example just straight up override callbacks etc and all they require is to be included `on_attach()`. But I guess I'm not 100% sure that all users will want all the features this will hopefully have or not.

I also am watching [this](https://github.com/neovim/nvim-lsp/issues/200) and seeing what happens. There is the chance that if the entire install goes away, it might make sense to just use Coursier here and add in the install/update/uninstall functionality. If we do that, then the next question is do we just handle the entire setup and not require nvim-lsp at all. /shrug

lua/metals.lua
landerlo: I've added the name_opt as well, so if any of those are passed it won't be requested in the UI. Passing the name is useful for the new symbol quickfix when we create the symbol under cursor.
The directory is probably less useful but it's still an option in the spec. In most cases people will just cd, but I can see -myself if I hace time- using a selection for all the existing packages and selecting destination. Maybe even using a full name: org.corp.NewClass that fills both the folder and the name into one call. Anyway, it doesn't add much code to support it. But feel free to remove if you think it's goldplatting.

lua/metals/util.lua
landerlo: This is my first lua code so I cannot answer.Yes, adding it to the module we can call it from outside util. Also if make it local the linter will complain that it isn't used, if it's exported in the module there is no warning even if not used anywhere.

landerlo:
If there is no update we can have a single nvim_setup that does everything, and user call it, and can  override bindings or Commands after calling it. Maybe that f with all batteries included can be calling three funs for more fine grained like setup() (does the attach), default_bindings(), and default_commands() so somebody can pick and choose to not have any unwanted bindings. I personally like to have new functionality and bindings added with as little config as possible, but offering the pick and choose possibility. For me the test is that a feature like this shouldn't require me any changes if I've gone batteries included. NewScalaFile appears magically when I PlugUpdate
 nvim-metals

lua/metals.lua
```diff
@@ -113,7 +170,7 @@ M.root_pattern = function(...)
 end
 
 -- Callback function to handle `metals/status`
--- This simply sets a global variable `metals_status` which can be easily
+-- This simply sets a global variable `meplugin/metals.vimtals_status` which can be easily
```
ckipp01, line 173: Was this changed by accident?

landerlo: Yes, I've reverted this,

lua/metals.lua
```diff
@@ -63,6 +63,63 @@ M.sources_scan = function()
...
+  end
+  if name_opt then
+    table.insert(args_string_array, 2, name_opt)
+  else
+    table.insert(args_string_array, 2, vim.NIL)
+  end
+
+  execute_command({
+      command = 'metals.new-scala-file';
+      arguments = args_string_array
```
ckipp01, line 119: I didn't realize this before, but on the Metals side, this is a bit odd to me that it's an array of two optional strings rather than 
```ts
{
  uri?: string,
  filename?: string
}
```

lua/metals/util.lua
ckipp01: No worries. This is fine as is, I was just curious.

ckipp01:
> If there is no update we can have a single nvim_setup that does everything, and user call it, and can override bindings or Commands after calling it. Maybe that f with all batteries included can be calling three funs for more fine grained like setup() (does the attach), default_bindings(), and default_commands() so somebody can pick and choose to not have any unwanted bindings. I personally like to have new functionality and bindings added with as little config as possible, but offering the pick and choose possibility. For me the test is that a feature like this shouldn't require me any changes if I've gone batteries included. NewScalaFile appears magically when I PlugUpdate
> nvim-metals

I fully agree with this. I think it makes sense to merge this is, and then start moving towards a setup like that.

PR #20, opened 2020-05-28
Call the correct execute_command function

timokau:
I have no lua experience, so maybe this is not actually the right thing
to do. The current implementation breaks with "attempt to call field"
though, and to me it looks like `execute_command` does not come from
`M`.


timokau:
Thanks for writing this plugin and for the quick reviews :)

Issue #4, opened 2020-05-02
Support the Decoration Protocol


ckipp01:
**Describe the feature**

Metals can publish worksheet evaluation using the [Decoration Protocol](https://scalameta.org/metals/docs/editors/decoration-protocol.html). Nvim has support for this using virtual text, and this is implemented in [coc-metals](https://github.com/scalameta/coc-metals/blob/master/src/decoration.ts) and it wasn't a ton of work.

**Search terms**

decoration protocol, worksheets

landerlo:
Have you started working on this?

ckipp01:
Hey @landerlo! Not yet, so if you want to give it a go, have at it.

ckipp01:
Btw, one thing to also note with this is that you'll want to also implement [`metals/didFocusTextDocument`](https://scalameta.org/metals/docs/editors/new-editor.html#metalsdidfocustextdocument) with this. There has been issues in a couple clients with the worksheets not being updated after the initial time, and it was due to this.

ckipp01:
Hey @landerlo, were you still planning on tackling this? If not, I have some free time and was thinking about giving it a go.

landerlo:
Hi, yes, I have something almost ready, I think I'll have some time to polish the PR tomorrow evening. I was testing it last night but I was having some issues with didFocus, metals doesn't seem to like it, and triggers some stack overflows. So I'll split that functionality to another PR, I don't think it's relevant to us, because what's needed for worksheets is to send didSave, nvim-metals only sends didChange at the moment, which works for normal scala files, but for worksheets, that's not triggering the decoration publication. DidSave notification is required in order to the worksheet to update. Maybe a bug in metals but we can workaround that by configuring the options to save.

landerlo:
PR here https://github.com/scalameta/nvim-metals/pull/22, 
It had a few headscratchers about some metals behaviour. First, it doesn't seem to send decorations on change, it only updates them on didFocus or didSave. Now only didSave is supported, which is fine by me, I want the reevaluation to happen on save anyway.
Another odd thing is that the httpEnabled option needs to be set, otherwise decorations don't work. I'd have thought that it had a false default. I'll send didFocus on additional PR but metals seems to have internal stack overflows when it receives the notification.

ckipp01:
Awesome, I'll start taking a look at it. Couple first thoughts:

[...]

PR #22, opened 2020-06-16
Support the Decoration Protocol #4

landerlo:
Instead of taking the color from metals options the color is configurable.

It doesn't include the hover messages. I've created a separate issue to discuss this behaviour.

### DIFF
--- a/lua/metals.lua
+++ b/lua/metals.lua
@@ -5,10 +5,7 @@ local util = require'metals.util'
 
 local M = {}
 
-M.decoration = {
-  namespace = api.nvim_create_namespace('metals_decoration');
-  color     = 'conceal';
-}
+local decoration_namespace = api.nvim_create_namespace('metals_decoration');
 
 local function execute_command(command, callback)
   vim.lsp.buf_request(0, 'workspace/executeCommand', command, function(err, method, resp)
@@ -202,7 +199,7 @@ end
 
 M['metals/publishDecorations'] = function(err, _, decorations)
   if err then
-    print("metals.publishDecorations: Server error:", err)
+    print("metals.publishDecorations: Server error")
   end
   if not decorations then
     return
@@ -221,10 +218,12 @@ M['metals/publishDecorations'] = function(err, _, decorations)
     return
   end
 
-  api.nvim_buf_clear_namespace(bufnr, M.decoration.namespace, 0, -1)
+  local decoration_color = vim.g.metals_decoration_color
+
+  api.nvim_buf_clear_namespace(bufnr, decoration_namespace, 0, -1)
 
   for _, deco in ipairs(decorations.options) do
-    util.set_decoration(bufnr, M.decoration.namespace, deco, M.decoration.color)
+    util.set_decoration(bufnr, decoration_namespace, deco, decoration_color)
   end
 end
 
--- a/lua/metals.lua
+++ b/lua/metals.lua
@@ -1,8 +1,14 @@
 local lsp = vim.lsp
 local vim = vim
+local api = vim.api
 local util = require'metals.util'
 
 local M = {}
+
+M.decoration = {
+  namespace = api.nvim_create_namespace('metals_decoration');
+  color     = 'conceal';
+}
 
 local function execute_command(command, callback)
   vim.lsp.buf_request(0, 'workspace/executeCommand', command, function(err, method, resp)
@@ -191,6 +197,34 @@ M['metals/status'] = function(_, _, params)
     vim.api.nvim_set_var('metals_status', '')
   else
     vim.api.nvim_set_var('metals_status', params.text)
+  end
+end
+
+M['metals/publishDecorations'] = function(err, _, decorations)
+  if err then
+    print("metals.publishDecorations: Server error:", err)
+  end
+  if not decorations then
+    return
+  end
+
+  local uri = decorations.uri
+  local bufnr = vim.uri_to_bufnr(uri)
+  if not bufnr then
+    print("metals.publishDecorations: Couldn't find buffer for ", uri)
+    return
+  end
+
+  -- Unloaded buffers should not handle diagnostics.
+  --    When the buffer is loaded, we'll call on_attach, which sends textDocument/didOpen.
+  if not api.nvim_buf_is_loaded(bufnr) then
+    return
+  end
+
+  api.nvim_buf_clear_namespace(bufnr, M.decoration.namespace, 0, -1)
+
+  for _, deco in ipairs(decorations.options) do
+    util.set_decoration(bufnr, M.decoration.namespace, deco, M.decoration.color)
   end
 end
 
--- a/lua/metals/setup.lua
+++ b/lua/metals/setup.lua
@@ -1,0 +1,13 @@
+--[[
+  Facilities to help modularize the metals plugin config
+--]]
+
+local M = {}
+
+--[[ The only way to enable didSave with current LSP and metals
+capabilities is to override after init. --]]
+M.on_init = function(client, _)
+  client.resolved_capabilities.text_document_save = true
+end
+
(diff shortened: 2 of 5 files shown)


lua/metals.lua
```diff
@@ -1,9 +1,15 @@
 local lsp = vim.lsp
 local vim = vim
+local api = vim.api
 local util = require'metals.util'
 
 local M = {}
 
+M.decoration = {
```
ckipp01, line 8: I don't love that this exposes `namespace` to the outside world when it probably shouldn't be. Another option could be to instead set something like:

```vim
let g:metals_decoration_color = 'Conceal'
```

Then in `plugin/metals.vim` do a check for it or assign it. Something like is done here: https://github.com/haorenW1025/completion-nvim/blob/2f579083a3585dffbf799ccb8bd5d89b175ca761/plugin/completion.vim#L8-L10

That way we wouldn't expose this entire thing. What do you think?

lua/metals.lua
```diff
@@ -182,4 +188,32 @@ M['metals/status'] = function(_, _, params)
   end
 end
 
+M['metals/publishDecorations'] = function(err, _, decorations)
+  if err then
+    print("metals.publishDecorations: Server error:", err)
```
ckipp01, line 193: We probably don't want to print the entire error do we? I think it may be best to just let the user know there is an error, and then let them to a `:MetalsToggleLogs` to get a closer look at what's going on. I'm just afraid the `err` here may not be then nicest thing to show there. What do you think?

lua/metals.lua
```diff
@@ -182,4 +188,32 @@ M['metals/status'] = function(_, _, params)
...
+  end
+  if not decorations then
+    return
+  end
+
+  local uri = decorations.uri
+  local bufnr = vim.uri_to_bufnr(uri)
+  if not bufnr then
+    print("metals.publishDecorations: Couldn't find buffer for ", uri)
```
ckipp01, line 202: Do you know if there is a way to color these? It might be good to get an `error`, `warning`, `info` in here to follow what the LSP log messages do. This doesn't need to be done for this, but just a note.

lua/metals/setup.lua
```diff
@@ -0,0 +1,13 @@
+--[[
+  Facilities to help modularize the metals plugin config
+--]]
+
+local M = {}
+
+--[[ The only way to enable didSave with current LSP and metals
+capabilities is to override after init. --]]
+M.on_init = function(client, _)
```
ckipp01, line 9: hmmm, not needed for this pr, but is this `on_init` something we could also add in all the `initializationOptions` that we are overriding?

nvim-lsp.vim
```diff
@@ -31,16 +31,20 @@ nnoremap <silent> <leader>ca  <cmd>lua vim.lsp.buf.code_action()<CR>
...
 " lua callbacks
 "-----------------------------------------------------------------------------
 :lua << EOF
   local nvim_lsp = require'nvim_lsp'
   local metals   = require'metals'
+  local setup    = require'metals.setup'
   local M        = {}
 
+  -- Decoration color. Choose from :highlight
+  metals.decoration.color = 'Conceal'
```
ckipp01, line 46: See my note above, but I think it's better to instead expose this as a global variable instead of through the lua module.

nvim-lsp.vim
```diff
@@ -56,13 +60,20 @@ let g:metals_server_version = '0.9.0+18-27d4652a-SNAPSHOT'
       inputBoxProvider             = true;
       quickPickProvider            = true;
       executeClientCommandProvider = true;
+      hoverProvider                = true;
```
ckipp01, line 63: What is this doing? There is no `hoverProvider` in `InitializationOptions` in Metals.

https://github.com/scalameta/metals/blob/master/metals/src/main/scala/scala/meta/internal/metals/InitializationOptions.scala

nvim-lsp.vim
```diff
@@ -56,13 +60,20 @@ let g:metals_server_version = '0.9.0+18-27d4652a-SNAPSHOT'
...
+      decorationProvider           = true;
+      isHttpEnabled                = false; -- Needs to be present
```
ckipp01, line 65: Yea, I'll look into Metals to see what is happening here. I think I actually may have an idea.

lua/metals/setup.lua
landerlo: I didn't want to change too much here, but moving forward it'll depend on the config strategy. IMO we should hide this from the user as we talked before and accept some overrides from them, so the setup function internals are hidden from the user. IMO, nvim-lsp doesn't even add much value for nvim-metals, and we could have a simply add nvim-metals and works with defaults or has a single call to setup_metals, accepting overrides from the user. Then the nvim-lsp setup function is left for DYI config. That would allow more freedom to change things in nvim-metals, though this is a topic for further discussions.

nvim-lsp.vim
landerlo: Strange, I must have copied it from an old config, hoverProvided used to exist, but it's gone from metals for a long time. Deleting

lua/metals.lua
landerlo: Yes, unifying the error mgt is pending, I don't know what are the best practices in nvim. In this case it's definitely worth removing the print as it prints the stackoverflow.

ckipp01: Sounds good 👍

lua/metals/setup.lua
ckipp01: > hide this away from the user

If we went this route and actually put all the `initializationOptions` in here as well, that would _hide it away_ though right?

And I get what you're saying about the value of `nvim-lsp`. At this moment it basically boils down to me needing to fully understand how to start and manage the client. Once I get there, I'll go that route and just manage installation and everything probably in this plugin.

nvim-lsp.vim
ckipp01: Well in the process of looking into this I found another issue. So that's good I guess 😏. We have that we are a `executeClientCommandProvider`, but don't actually have all the cases for all the commands. However, I think that's fine as clients should be able to support a subset of these. The pr in metals firstly just make sure that doctor still works in a client has `isHttpEnabled = true` and also `executeClientCommandProvider = true`.

https://github.com/scalameta/metals/pull/1834

landerlo: Yeah, the http service works fine when enabled. The executeClientCommandProvider was there already from the new file PR. This is initiated from the client, so we can support any subset we want.

ckipp01: Alright, I found the issue with Doctor. Once pr I have linked above is merged in, we can go ahead and remove the `isHttpEnabled = false`.

Also, you have a merge conflict since I did add some stuff earlier today.

landerlo: Ok, I've pushed the rebased PR

landerlo: BTW, which font are you using for the g:diagnostic_virtual_text_prefix character in nvim-metals.vim?

ckipp01: I actually don't have `g:diagnostic_enable_virtual_text` turned on. When I did I was just using `` as the prefix.

nvim-lsp.vim
```diff
@@ -56,13 +60,21 @@ let g:metals_server_version = '0.9.0+236-8d0924af-SNAPSHOT'
       inputBoxProvider             = true;
       quickPickProvider            = true;
       executeClientCommandProvider = true;
+      decorationProvider           = true;
+      --[[ isHttpEnabled needs to be explicitly set to true or false
+         otherwise there is an issue with metals ]]--
+      isHttpEnabled                = false; -- Property needs to be explicitly set
```
ckipp01, line 66: Alright, the pr was merged, so this should now work fine with this all removed.

PR #23, opened 2020-06-16
Did focus provider

landerlo:
This adds did focus but I see metals throwing exceptions when a notification is sent, the server recovers so we can simply remove the error printing in nvim-metals and merge, though at the moment I don't see the focus refresh being useful, when save event works for evaluations. 

It is rebased on top of decoration_provider PR

### DIFF
--- a/lua/metals.lua
+++ b/lua/metals.lua
@@ -227,4 +227,15 @@ M['metals/publishDecorations'] = function(err, _, decorations)
   end
 end
 
+-- Notify the server when document has been focused
+-- This needs to be called in the appropriate autocommand, i.e. FocusGained
+M.did_focus = function()
+  local focused_uri = vim.uri_from_bufnr(0)
+  vim.lsp.buf_request(0, 'metals/didFocusTextDocument', focused_uri, function(err, _, _)
+    if err then
+      print('metals/didFocusTextDocument: Server Error')
+    end
+  end)
+end
+
 return M
--- a/lua/metals/setup.lua
+++ b/lua/metals/setup.lua
@@ -10,4 +10,10 @@ M.on_init = function(client, _)
   client.resolved_capabilities.text_document_save = true
 end
 
+M.auto_commands = function()
+  vim.api.nvim_command [[augroup NvimMetals]]
+    vim.api.nvim_command [[autocmd BufEnter <buffer> lua require'metals'.did_focus()]]
+  vim.api.nvim_command [[augroup end]]
+end
+
 return M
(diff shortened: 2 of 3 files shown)


landerlo:
@ckipp01 I saw you did some work on the metals and lsp4j integration. Maybe you can figure out what's going on, though it might be some undocumented behaviour required about this callback?
I haven't did deeper but the sending of didSave as in the decoration PR is preferable to me.

landerlo:
Rebased, I see what you meant about didFocus. Testing with the current master the isHttp issue went away but the opening various files needs the refocus refresh to get things updating again.
But that causes a metals exception , though it seems recoverable.

landerlo:
I've changed the autocmd to BufEnter instead of FocusGained and now the issue with the loss of focus goes away, but the server errors keep happening.

landerlo:
Maybe we can disable the error print as metals keeps responding and worksheets work fine, though the metals logs will keep increasing so that's probably a merge stopper. I wonder if it has to do with message ordering.

ckipp01:
> Maybe we can disable the error print as metals keeps responding and worksheets work fine, though the metals logs will keep increasing so that's probably a merge stopper. I wonder if it has to do with message ordering.

Hmmm, yea I see what you mean now. I'll try to spend some time today comparing the logs with coc-metals and nvim-metals with worksheets to see if I can track down where the exception is coming from. I'd like to get that figured out before we merge this.

landerlo:
coc implements slow-task, which I do not. It should have a fallback, but maybe that's why is  failing. Because the stackoverflow happens inside the parsing code.

ckipp01:
Hey! So I'm still working on chasing this down. Something related to the `metals/didFocusTextDocument` is causing it because I can finally reproduce this consistently and accurately.

lua/metals.lua
```diff
@@ -227,4 +227,15 @@ M['metals/publishDecorations'] = function(err, _, decorations)
   end
 end
 
+-- Notify the server when document has been focused
+-- This needs to be called in the appropriate autocommand, i.e. FocusGained
+M.did_focus = function()
+  local focused_uri = vim.uri_from_bufnr(0)
+  vim.lsp.buf_request(0, 'metals/didFocusTextDocument', focused_uri, function(err, _, _)
```
ckipp01, line 234: Are you ready for this??

This is sending the `didFocusTextDocument` *request* to the server, and this needs to be a notify. Changing this to:

```suggestion
  vim.lsp.buf_notify(0, 'metals/didFocusTextDocument', focused_uri, function(err, _, _)
```
fixes the issue.

landerlo: Wow! TIL about notify, thank you for digging into that, I was sure it had to do with the interaction, but I didn't know there was another pattern.

landerlo: Give me bigger :man_facepalming:

Issue #31, opened 2020-11-21
Replace our full workspace diagnostics feature with the built-in one.


ckipp01:
**Task**

We should be able to use the newly added https://github.com/neovim/neovim/pull/13285 to get all diagnostics instead of per buffer like they exposed before. This can replace the hacky way we do this now.

**Search terms**

workspace diagnostics

landerlo:
I have a PR that does this, it's a simple change, just waiting for https://github.com/neovim/neovim/pull/13310 to be merged so I can test. It puts all the errors on top, to be more aware of them. Maybe this should be more configurable, in sane codebases where the warnings are treated the same as errors you might want to have them together, but as a list of all diagnostics I find more useful to tackle errors first and see the errors first. Current codebase I'm working on has too many Warnings.

 By hacky way I imagine you refer to https://github.com/scalameta/nvim-metals/issues/30, the refresh issue. I commented on that one. We should keep this too separate.

ckipp01:
Going to go ahead and close this now that #32 is merged. Thanks again for addressing this @landerlo!

> By hacky way I imagine you refer to #30, the refresh issue. I commented on that one. We should keep this two separate.

Indeed, honestly I when I wrote this and first briefly saw the new diagnostic stuff I didn't look as closely as I should have and just shot this in here to not forget about it. I probably shouldn't have used the word hacky, as I didn't mean anything negative by it.

Cheers.

landerlo:
No worries, I didn't read it negatively. It isn't ideal as it is now so it's good to keep the issue open and we'll have to tackle it at some point, though I'm quite happy working with the all diagnostics now, makes it usable for serious development.

PR #32, opened 2020-12-03
#31 GetAll diagnostics as per new LSP

landerlo:
PR for https://github.com/scalameta/nvim-metals/issues/31


ckipp01:
Thanks for sending this in @landerlo! I plan to review this later this week. I haven't really been keeping up with the LSP changes in nvim, and there seems to be a lot that has changed, and my entire setup is now broke. So I want to take a bit of time to really get caught up, and then I'll properly review this.

ckipp01 (approved):
Actually, after looking through this a bit, there is no reason to not just merge this right away since I see that `lsp.util.diagnostics_by_buf` is throwing a warning. Again, thanks for a taking a look at this. LGTM!

PR #37, opened 2020-12-22
Remove the need for nvim/nvim-lspconfig

ckipp01:
I still need to clean up a few things and pretty much replace all of the documentation/readme, but this is the approach that I'd like to take. To summarize, this pr will get rid of the need for a user to have `nvim/lsp-config` installed. They still can if they want to, but this should become the preferred way to use Metals with nvim lsp mainly because of all the extra functionality and Metals-specific lsp extensions it provides. The main idea is that instead of having a user add in a bunch of handlers on their own and also have to keep adding them in as we add them, is for them to simply have the following:

```lua
if has('nvim-0.5')
  augroup lsp
    au!
    au FileType scala lua require('metals').initialize_or_attach({})
  augroup end
endif
```

This will basically provide _all_ of the functionality that `nvim/nvim-lspconfig` and this extension provided before, but with way less work to setup. However, this still remains very customizable.  For example, my personal setup locally looks like this:

```lua
:lua << EOF
  metals_config = {}
  metals_config.handlers = {}
  metals_config.init_options = {}

  metals_config.on_attach = function()
    require'completion'.on_attach();
  end

  metals_config.init_options.statusBarProvider = 'on'

  metals_config.handlers["textDocument/publishDiagnostics"] = vim.lsp.with(
    vim.lsp.diagnostic.on_publish_diagnostics, {
      virtual_text = {
        prefix = '',
      }
    }
  )
EOF

if has('nvim-0.5')
  augroup lsp
    au!
    au FileType scala lua require('metals').initialize_or_attach(metals_config)
  augroup end
endif
```

This allows me to use the `on_attatch()` from nvim-completions, change the default `show-message` to `on` to be able to use `metals/status`, and also customize how my diagnostics look.

This plugin also will fully handle the installation of Metals for the user by providing a `:MetalsInstall` command to put Metals in a specific location and use it. I've tried to be quite descriptive with messages to help the user know what's going on.

I still have some further things to do, but this will at least give an idea of the direction I'm headed.

@landerlo feel free to take a peak at this and give me any feedback.

TODO
- [x] Update readme with new instructions (keep in short, and focus on the help docs instead)
- [x] Update the metals.txt with basically all the information a user would need
- [x] Add in a `:MetalsInfo` command to give some practical info, basically the output of `metals --version` which shows the version of metals and also the supported Scala Versions. This command can also be used to show where Metals is installed
- [x] Fix CI -- figure out if I just want to use lua-format instead of luacheck
- [x] Figure out what I should do with the `set shortmess-=F`. This _needs_ to be set in order for the FileType autocmd to display `print`, which is needed for the helpful messages.


ckipp01:
For anyone that stumbles onto this, you can find a short migration guide for these changes here: https://github.com/scalameta/nvim-metals/discussions/17#discussioncomment-238344

PR #74, opened 2021-01-16
Change install to use vim.loop.spawn.

### LINKED ISSUE #65: Add some sort of progress to show that `:MetalsInstall` is actually doing something.
**Describe the feature**

When a user uses `:MetalsInstall`, it's really unclear that anything is going on. It just blocks until it's done. It'd be great to show some time of progress status while that is happening.

Maybe re-use the Metals status.

**Search terms**

Install progress

### LINKED ISSUE #71: Replace `os.execute` call with `vim.loop.spawn` for longer coursier/metals calls.
**Task**

Currently things like `:MetalsInstall` and ~`:MetalsInfo`~ (I don't actually think this one is needed) block since they take a bit of time. Instead of using `os.execute` and blocking we can swap those out to use `vim.loop.spawn` to do them in the background. I think this will be a better user experience.

ckipp01:
This change ensures that when a user is using `:MetalsInstall` it doesn't block.

Another addition is adding in  https://github.com/tjdevries/vlog.nvim which introduces
nvim-metals specific logging. I plan to actually overhaul this a bit for my needs, but
this just introduces it.

Closes #71
Closes #65

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -1,6 +1,12 @@
 -- vim ft=lua tw=80
 
 max_line_length = 100 
+globals = {
+  "Coursier_handle"
+}
 read_globals = {
   "vim",
+}
+exclude_files = {
+  "lua/metals/log.lua"
 }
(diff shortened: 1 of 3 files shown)


ckipp01:
Closes #71 
Closes #65

PR #88, opened 2021-01-20
Set buffer-local variable to identify metals logs buffer

### LINKED ISSUE #87: MetalsLogsToggle is confused if another terminal is open
When you have an integrated terminal open in nvim, running MetalsLogsToggle will say that logs are already opened.

1. Open any metals project
2. Open an integrated terminal
3. Run :MetalsLogsToggle

**Expected behavior**

A terminal is opened tailing metals logs

**Actual behavior**

Command things logs are already opened

**Installation:**

- Operating system: Linux
- Nvim version: v0.5.0-dev+1012-gfe1ebea33
- Commit of nvim-metals: 13635adae2d4a64d41d48ed8587ee022825e2e17
- Metals version: 0.9.10

**Search terms**

metals, logs, terminal, nvim, toggle

keynmol:
Fixes #87 

~I haven't installed everything from CONTRIBUTING.md, just~ want to check if approach makes sense :) 

I'm doing it this way because I couldn't find an API that parses `term://{cwd}//{pid}:{cmd}` format that is in buffer name for terminals.

If there's such an API (and if it's preferable to use it), let me know @ckipp01

### DIFF
--- a/lua/metals.lua
+++ b/lua/metals.lua
@@ -129,15 +129,26 @@ end
 
 M.logs_toggle = function()
   local bufs = api.nvim_list_bufs()
-  for _, v in ipairs(bufs) do
-    local buftype = api.nvim_buf_get_option(v, 'buftype')
-    if buftype == 'terminal' then
-      log.info_and_show('Logs are already opened. Try an :ls to see where it is.')
+
+  for _, buf in ipairs(bufs) do
+    local buftype = api.nvim_buf_get_option(buf, 'buftype')
+    local _, purpose = pcall(api.nvim_buf_get_var, buf, 'metals_buf_purpose')
+
+    if buftype == 'terminal' and purpose == 'logs' then
+      local first_window_id = vim.fn.win_findbuf(buf)[1]
+      if first_window_id then
+        vim.fn.win_gotoid(first_window_id)
+      else
+        api.nvim_command(string.format('vsp | :b %i', buf))
+      end
+
       return
     end
   end
+
   -- Only open them if a terminal isn't already open
   api.nvim_command [[vsp term://tail -f .metals/metals.log]]
+  vim.b['metals_buf_purpose'] = 'logs'
 end
 
 -- Implements the new-scala-file feature.
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -2,10 +2,8 @@
 
 max_line_length = 100 
 globals = {
-  "Coursier_handle"
-}
-read_globals = {
-  "vim",
+  "Coursier_handle",
+  "vim"
 }
 exclude_files = {
   "lua/metals/log.lua"


ckipp01:
Thanks for the pr @keynmol! So I just played around with this a bit, and it seems that it's allowing me to open multiple.

Screenshot 2021-01-20 at 17 48 57

My guess would be that this: `| :let b:metals_buf_purpose="logs"` isn't working as expected. One thing I also thought of was instead of warning that one is already open, and telling to user to find and open it. What if we just opened it? Since we know that a user wants to open it if they are doing a `:MetalsLogToggle`. I think that'd be cleaner. So the flow would be:

- User opens and closes logs
- User does another `:MetalsLogsToggle`
- It detects that there is a buffer for the logs already, splits and opens the buffer there.

What do you think?

keynmol:
> What do you think?

Yeah, from UX point of view that's what I would expect.

Strange that it doesn't work for you - I will test it again, but I was 100% certain that my tests showed that it helped. 

I'll see what's up. And if you do `:echo b:metals_buf_purpose` (or what the syntax is to print out buffer variable?) in either of those terminals, does it show?

ckipp01:
> > What do you think?
> 
> Yeah, from UX point of view that's what I would expect - I'll fix it
> 
> Strange that it doesn't work for you - I will test it again, but I was 100% certain that my tests showed that it helped.
> 
> I'll see what's up. And if you do `:echo b:metals_buf_purpose` (or what the syntax is to print out buffer variable?) in either of those terminals, does it show?

Ahh, actually I see what's happening. It does correctly set "logs", but when you are in the for loop, the `vim.b` check is checking the local buffer that you issued the command from, not the buffer that you're iterating over.

keynmol:
Oh right. 

I'll use nvim's API to get any buffer's var. I had it before but couldn't figure out the idiomatic way to wrap it in pcall so it would stop throwing :)

lua/metals.lua
```diff
@@ -129,15 +129,16 @@ end
...
-      log.info_and_show('Logs are already opened. Try an :ls to see where it is.')
+    if buftype == 'terminal' and vim.b.metals_buf_purpose == 'logs' then
+      log.info_and_show(string.format('Logs are already opened. Try an :ls to see where it is.' ..
+                                          'Or switch to it directly using :b%i', i))
       return
     end
   end
   -- Only open them if a terminal isn't already open
-  api.nvim_command [[vsp term://tail -f .metals/metals.log]]
+  api.nvim_command [[vsp term://tail -f .metals/metals.log | :let b:metals_buf_purpose="logs"]]
```
ckipp01, line 141: Also this is a bit of a nit, but I try to avoid doing to much in a command like this, so we can also separate this out.
```suggestion
  api.nvim_command [[vsp term://tail -f .metals/metals.log]]
  vim.b.metals_buf_purpose = 'logs'
```

keynmol:
```
    lua/metals.lua:145:3: setting read-only field b.metals_buf_purpose of global vim
```

🙃 

Is there a way to make this check more fine-grained?

ckipp01: You an just add it to `globals` in the `.luacheckrc` file.

keynmol:
> It detects that there is a buffer for the logs already, splits and opens the buffer there.

Just took a look at this - what is the best way to detect that the buffer is "open", i.e. part of the window?

I figured that while I can find that there's a matching buffer, just doing `:vsp` on that buffer would lead to the buffer being opened all the time, right?

So there needs to be a way to "go" to a visible window/tab that contains this buffer?

ckipp01:
> > It detects that there is a buffer for the logs already, splits and opens the buffer there.
> 
> Just took a look at this - what is the best way to detect that the buffer is "open", i.e. part of the window?

I'm not actually sure, I may have to dig a bit. I think maybe `nvim_buf_is_loaded()`?

> I figured that while I can find that there's a matching buffer, just doing `:vsp` on that buffer would lead to the buffer being opened all the time, right?

Well if you do a `:vsp #` on the buffer you are targeting it will just split again with that buffer as yet another window, so you don't want that.

> So there needs to be a way to "go" to a visible window/tab that contains this buffer?

Again, without really digging in I'm not sure, but I think maybe `win_gotoid()` may do what you want.

keynmol:
Does this look better?

https://user-images.githubusercontent.com/1052965/105226302-b9a7db00-5b57-11eb-86fb-17cf1b88bf35.mov

ckipp01:
> Does this look better?
> Screen.Recording.2021-01-20.at.19.41.17.mov

hey! yes! looking better! However, I may have lead you astray a little bit because this situation works when working tabs like you show, but no so much if you open the buffer with the term, close the buffer (but the buffer still exists), and then you try to toggle again. Because it will find the `term` in the for loop, but `win_findbuf(buf)[1]` will be `nil` since it's a buffer not attached to a window at the moment. Son in that scenario, you found the buffer, but it's not in a window, so you should be able to then just split where you are at and open that buffer in the split. Here is a vid of me showing that workflow, where I still end up with 2 log terms.


https://user-images.githubusercontent.com/13974112/105321632-fdd2c400-5bc7-11eb-93f2-611eaf2171bd.mov

keynmol:
Aand the recording is too big.

The new logic is to vsplit to the found buffer if the window is not found

ckipp01:
> Aand the recording is too big.
> 
> The new logic is to vsplit to the found buffer if the window is not found

Haha, no worries, I just tested it and it works as expected! Thanks a lot for tackling this!! Are you able to squash your commits down, and then LGTM!

ckipp01 (approved):

PR #91, opened 2021-01-29
Add in fallbackScalaVersion setting.

ckipp01:
This is a setting introduced here: https://github.com/scalameta/metals/pull/2417


ckipp01:
Just for anyone that may stumble on this and wonder why it may not be working, it requires the most recent snapshot of Metals. So if 0.9.11 isn't out yet, that's why.

PR #103, opened 2021-02-18
Basic nvim-dap integration

### LINKED ISSUE #57: Provide integration with `mfussenegger/nvim-dap`
**Describe the feature**

A [nvim-dap](https://github.com/mfussenegger/nvim-dap) integration would be awesome.

We'd have to change some of the `init_options` when we start Metals to make sure that `debuggingProvider` is set to `true`. We'd also need to figure out the configuration since normally in metals the `run` and `debug` are always triggered via code lenses. However, we should be able to work around that.

**Search terms**

debugging, nvim-dap

ckipp01:
~This introduces three new basic commands. `:MetalsRun`, `:MetalsTestFile`, and `:MetalsTestTarget`. The majority of the changes needed to make this work are actual in the server itself. I'll have a pr open for that soon, but this is the base of how I'm testing it.~

See the comments down below for updates

Closes #57

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,10 +9,9 @@ necessary commands you'll need to develop with nvim and Metals. This extension
 also implements many of the custom Metals LSP extensions that will give you a
 much richer experience than just using Metals with the default
 [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig) setup, as well as
-automatically setting all of the correct `init_options`.
-
-_NOTE:_ Keep in mind that the level of support is rapidly changing, so expect
-stuff to change without warning.
+automatically setting all of the correct `init_options`, and offering some
+integrations with other projects such as
+[nvim-dap](https://github.com/mfussenegger/nvim-dap).
 
 <p align="center">
     <a href="https://gitter.im/scalameta/metals-vim">
@@ -76,3 +75,9 @@ help docs. Similarly, to see the available configuration options, check out
 _NOTE_: You can find an example of a minimal configuration showing how to set
 various settings and options
 [here](https://github.com/scalameta/nvim-metals/discussions/39).
+
+## Integrations
+
+To see the full details on the available configurations, checkout out `:help
+metals-integrations`. The currently available integrations are:
+- [nvim-dap](https://github.com/mfussenegger/nvim-dap)
--- a/doc/metals.txt
+++ b/doc/metals.txt
@@ -16,9 +16,11 @@ CONTENTS                                                           *nvim-metals*
     6. Lua API............... |metals-lua-api|
     7. Custom Handlers....... |metals-custom-handlers|
     8. Functions............. |metals-functions|
+    9. Integrations...........|metals-integrations|
 
 ================================================================================
 INTRODUCTION                                                *metals-introdction*
+
 nvim-metals is a Lua plugin built to provide a better experience while using
 Metals, the Scala Language Server, with Neovim's built-in LSP support. This
 plugin provides the necessary commands you'll need to develop with nvim and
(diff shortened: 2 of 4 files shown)


alexcardell:
Super excited to try this when it's ready. Is it feasible to export the configuration and use nvim-dap's native lua API directly? To use existing mappings for starting runs/tests, etc.

ckipp01:
> Super excited to try this when it's ready. Is it feasible to export the configuration and use nvim-dap's native lua API directly? To use existing mappings for starting runs/tests, etc.

Good question, so this is _super_ rough at the moment, and this was just the easiest way for me to test as I was making changes to the server. So my initial thought is that you will just be able to definition a `dap.configurations.scala` locally if you want and the same with `dap.adapters.scala`. So you'll have a lot more control than what it looks like at the moment and it will merge configurations with what you have locally overwriting what the default is. The only tricky part about using the native nvim-dap lua api directly that for example you can't really use `continue()` until you know the host/port. Some of that is dynamic, so the `metals.debug-adapter-start` needs to be sent first to Metals to determine if we are doing a run, get the main class or classes that you can choose between, or if you are doing a single test of the current file or a testing the entire build target. It's first after that that metals sends back that info. So my plan is that you should still be able to use everything natively in nvim-dap, _except_ the process will probably have to be initiated with one of the new metals commands. I really want to make it as smooth as possible with the user needing to configure as little as possible. So in theory, it'd be great for the user to not have to set anything up in terms of configurations or adapters if they don't want to, they just need to know the nvim-dap commands for debugging.

I'm totally open to suggestions for all this, since it's all in flux at the moment. Up until this point if you didn't have code lenses with metals, you needed a json file to configure how to run or test, and I want to avoid that.

ckipp01:
Once https://github.com/scalameta/metals/pull/2532 is finished I'll start hammering out more details here.

mfussenegger (commented):
Cool to see more nvim-dap integrations. If I may, I think I've some inputs that might help

> So my initial thought is that you will just be able to definition a dap.configurations.scala locally if you want and the same with dap.adapters.scala. So you'll have a lot more control than what it looks like at the moment and it will merge configurations with what you have locally overwriting what the default is. The only tricky part about using the native nvim-dap lua api directly that for example you can't really use continue() until you know the host/port. 


I think there is a way to setup things so that the users can end up using `continue()` to select the configuration:

The adapter function receives the selected `configuration` (`:help dap-configuration`) as second argument. This would allow you to define an adapter like this:


```lua
-- you'll probably want to move this into a setup() method or something

local status, dap = pcall(require, 'dap')
if not status then
  log.error_and_show(
      'Unable to find nvim-dap. Please make sure mfussenegger/nvim-dap is installed.')
  return
end

dap.adapters.scala = function(callback, config)
  local uri = vim.uri_from_bufnr(0)
  -- `runType` is arbitrary here, I just used `runType` because that's what the execute_command argument is called. Might be worth prefixing or otherwise come up with a name that makes it clear that this is used by nvim-metals in the client, and not an option of the debug adapter itself
  local runType = config.runType or 'run'
  execute_command({
    command = 'metals.debug-adapter-start',
    arguments = {path = uri, runType = runType}
  }, function(_, _, res)

    local port = util.split_on(res.uri, ':')[3]

    -- You could additionally have a `enrich_config` function here that creates a copy of the config and removes all `nvim-metals` specific properties from it before calling `on_config` to ensure that the debug-adapter doesn't receive any information it can't handle
    callback({type = 'server', host = '127.0.0.1', port = port})
  end)
end
```

Note that now the adapter function executes the LSP command.
And then have different configurations:


```lua

dap.configurations.scale = {
  {
     type = 'scala',
     request = 'launch',
     name = 'Run file',
     console = 'integratedTerminal',
     runType = 'run' -- this is used by the adapter function
  },
  {
     type = 'scala',
     request = 'launch',
     name = 'Test file',
     console = 'integratedTerminal',
     runType = 'testFile' -- this is used by the adapter function
  },
}
```


An alternative would then also be to have specific functions. Some made up example:


```lua
M.dap_run_test_case = function()
  local test_case = some_logic()
  -- this is basically a "run once config" (but can be triggered again with dap.run_last())
  local config = {
    type = 'scala',
    request = 'launch',
    name = 'Run test case',
    ...
  }
  dap.run(config)
end
```

ckipp01:
Thanks a lot for swinging by @mfussenegger!! Ha, so yes, what I have right now is quite hacky as I was doing the very bare minimum to test some new stuff that I was adding to the server for main and test discovery, as right now we rely on the user passing in the exact class or test. I'm just about done with that, and then I'll dive back into this. Thanks a lot for you insight 🙏🏼. Once I dive back in I may throw some questions your way 😬 .

ckipp01:
So I ended up going with your first recommendation @mfussenegger. Thanks again for you input. I ended up going quite smoothly, and I like that I don't need to add any extra commands like I thought. If you have a moment, please take another look, but no pressure if you don't have time. I'll use this in the coming day or so, and then merge if it seems alright.

For anyone else following along, this adds basic integration with nvim-dap by just having a simple `on_attatch` that needs to be added when you want to use this. That will ensure that the correct adapter for scala is configured, and then you just need to provide the configurations you'd like. I gave an example in the help doc with the three run types that are available in Metals. Keep in mind that this was just merge into Metals, so it's not in a stable release yet, and to use it you'd need to use the snapshot.

mfussenegger:
Looks good 👍

The only other thought I had is that it may be easier for other DAP clients if the adapter start command wouldn't already have to choose what to execute, but if that were up to the configuration via the launch request.

But that would require changes in the debug adapter implementation. No clue how feasible that is.

PR #105, opened 2021-02-26
Pass client and bufnr parameter to user on_attach function

Kazy:
If you do `:help lsp.start_client`, you can find the definition of the `on_attach` callback, which takes two parameters. For instance when using `lsp-status`, one need to pass the parameters to `require('lsp-status').on_attach` for it to initialize properly.


ckipp01 (approved):
Thanks for the contribution @Kazy! Totally missed this, thanks for catching it! LGTM!

Issue #120, opened 2021-04-11
Don't register the nvim-metals commands unless `initialize_or_attatch` is called.


ckipp01:
**Task**

Right now the `:Metals*` commands are sort of bleeding into everything. This is the sort of thing I'd be annoyed at another plugin doing. Instead of just having them in the `plugin/metals.vim` file I should instead just move them into the attach functionality and first add them when the plugins is _actually_ going to be used.

**Additional context**

Honestly, once we remove these the only real things left in here are the `metals#status()`, `metals#errors()`, and `metals#warnings()` functions. I'd like to just remove this too, but should ensure that there are easy replacements for them and that people are notified since this will be a breaking change.

**Search terms**

Stop leaking your dirty nvim-metals command everywhere.

ckipp01:
Closed via https://github.com/scalameta/nvim-metals/pull/126

PR #124, opened 2021-04-12
Remove `:Format` command.

ckipp01:
This was sort of left over from when I just started the plugin and was
also creating extra LSP helper commands. This is the last one left that
I sort of just forgot about, and it probably just shouldn't be in here.
All it's doing is just calling `vim.lsp.buf.formatting()`, so it can
very easily be mapped locally. Plus most people may already have this
mapped directly calling that anyways.

Plus, is is also conflicting with another plugin:
  https://github.com/mhartington/formatter.nvim/issues/35

### DIFF
--- a/doc/metals.txt
+++ b/doc/metals.txt
@@ -367,7 +367,6 @@ handle it.
 COMMANDS                                                       *metals-commands*
 
 The following commands are provided by nvim-metals:
-  * |Format|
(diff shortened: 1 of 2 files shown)


ckipp01:
This also sort of relates to https://github.com/scalameta/nvim-metals/issues/120, which I'll take care of later this week.

PR #138, opened 2021-05-10
Introduce reliance on plenary

### LINKED ISSUE #125: Replace my UI stuff with Plenary
**Task**

So looking at the docs for`nvim_open_win()` the config can just take a border argument. There is no title like I currently have, but it might simplify things a lot, which I'm all about. I sort of hate all the stuff I have in ui.lua, and I wouldn't mind literally replacing all of it if I could

Related to this, something else that has been on my mind is that there are certain things in `nvim-metals` that also part of https://github.com/nvim-lua/plenary.nvim. I think if we are going to start ripping stuff out and replacing it, it may be a good time to really go through everything and see what we could remove via plenary, and then just require plenary as a dependency.

ckipp01:
ATM this is a WIP, but there are a fair amount of things that I think we can just rely on Plenary for. Since most users will have this installed anyways if they are using `ckipp01/scala-utils`, ` nvim-telescope/telescope.nvim`, etc then this probably isn't a huge deal. Plus it will allow for me to get rid of a bunch of code in here that is quite fragile.

- [x] Replace UI with `plenary.window`
- [ ] Replace path uitls with `plenary.path` (before doing this add tests for path utils)
- [ ] Update docs and examples to include plenary

This will close #125


ckipp01:
Superceded by https://github.com/scalameta/nvim-metals/pull/229

PR #143, opened 2021-05-29
Add support for the Tree View Protocol

### LINKED ISSUE #101: Support the tree view protocol
**Describe the feature**

One large thing missing in nvim-metals is support for the [Tree View Protocol](https://scalameta.org/metals/docs/integrations/tree-view-protocol.html). One of the most useful parts of this is being able to explore external sources and having an easy way to jump from the symbol you're on.

**Search terms**

tree view protocol

ckipp01:
This is pretty raw still at the moment, but I figured I'd push it up. Few things I need to do before being ready to merge:

- [x] Add some documentation
- [x] There is some duplication and unnecessary branches I can get rid of. Go through and simplify

While I'd like to get this done, it'd be a decent size refactor. I have the UI and the tree a bit too tightly coupled in my opinion. So really I need to thoughtfully decouple this and move it out.

- [x] See if it's possible to separate `Tree` out just like I did with `Node`.

I need to work on this, however I want to get this merged in and get some usage since for the most part, I don't _believe_ people will actually hit on this. Either way, it's what I'll probably work on next.

- [x] Figure out a plan for the caching of the tree to do a comparison so the nodes don't close when a parent is replaced

I'll create another ticket for this. I actually want this, but want to think a bit more about the best way to show this.
- [x] Do we want to differentiate type by color or anything?

Closes #101

### DIFF
--- a/doc/metals.txt
+++ b/doc/metals.txt
@@ -14,9 +14,10 @@ CONTENTS                                                           *nvim-metals*
     4. Options............... |metals-options|
     5. Commands.............. |metals-commands|
     6. Lua API............... |metals-lua-api|
-    7. Custom Handlers....... |metals-custom-handlers|
-    8. Functions............. |metals-functions|
-    9. Integrations...........|metals-integrations|
+    7. TVP Module............ |metals-tvp-module|
+    8. Custom Handlers....... |metals-custom-handlers|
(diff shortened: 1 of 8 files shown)


ckipp01:
Closes #101

PR #156, opened 2021-06-13
Add commands in to test code lenses

ckipp01:
Once https://github.com/neovim/neovim/pull/13165 is merged we can finish this up.


ckipp01:
Note that you'll need the latest HEAD of NeoVim so it has the following commit: https://github.com/neovim/neovim/pull/13165

Also I made an opinionated choice to put in the autcmd to refresh code lenses. I don't want to make users manually have to add another setting to enable this. Therefore it's on by default. They still need to enable the super method code lenses if they want those.

PR #169, opened 2021-07-04
Add in dependabot for github actions and add next to ci branches

ckipp01:


ckipp01:
@mergify backport next

PR #176, opened 2021-07-11
refactor metals.lua for new_scala_file

eugenemedvediev:
Refactor metals.lua:
- simplified M.new_scala_file
- used fn variable instead of vim.fn
- fixed new_scala_file documentation link
- fixed typo `Parameters`
- (optional) set parameters name for `execute_comand` in `M.super_method_hierarchy`

CONTRIBUTING.md
- fixed typo in CONTRIBUTING.md

### DIFF
--- a/lua/metals.lua
+++ b/lua/metals.lua
@@ -13,7 +13,7 @@ local util = require("metals.util")
 local M = {}
 
 -- General function used to execute various server commands.
--- @param command_params (optional, table) Paramets to send to the server (arguments and command).
+-- @param command_params (optional, table) Parameters to send to the server (arguments and command).
 -- @param callback (function) callback function for the request response.
 local function execute_command(command_params, callback)
   lsp.buf_request(0, "workspace/executeCommand", command_params, function(err, method, resp)
@@ -26,7 +26,7 @@ local function execute_command(command_params, callback)
 end
 
 M.analyze_stacktrace = function()
-  local trace = vim.fn.getreg("*")
+  local trace = fn.getreg("*")
   if trace:len() > 0 then
     execute_command({ command = "metals.analyze-stacktrace", arguments = { trace } })
   else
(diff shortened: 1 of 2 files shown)


lua/metals.lua
```diff
@@ -92,7 +92,10 @@ M.copy_worksheet_output = function()
         -- return a warning with logMessage, so we can skip it here.
       end
     end
-    execute_command({ command = "metals.copy-worksheet-output", arguments = uri }, copy_response)
+    execute_command({
+      command_params = { command = "metals.copy-worksheet-output", arguments = uri },
+      callback = copy_response,
+    })
```
ckipp01, line 98: Hey! So named params like this won't actually work. They work inside of the actual parameter argument since it's a table, so it expects that format. So this will have to be reverted.

lua/metals.lua
```diff
@@ -173,30 +176,21 @@ M.logs_toggle = function()
...
+  local args_string_array = {}
 
-  if file_type_opt then
-    table.insert(args_string_array, 3, file_type_opt)
-  else
-    table.insert(args_string_array, 3, vim.NIL)
-  end
+  table.insert(args_string_array, 1, directory_uri_opt)
+  table.insert(args_string_array, 2, name_opt)
+  table.insert(args_string_array, 3, file_type_opt)
```
ckipp01, line 193: 👍🏼 much cleaner. Although I hate on the metals side we have to send it in like this.

ckipp01 (changes requested):
Thanks for the pr @eugenemedvediev! We'll need to revert the named params back, but other than that this is good!

lua/metals.lua
eugenemedvediev: I found it here: https://www.lua.org/pil/5.3.html
Looks `lint` or `format` commands added extra () around {} which was not my intent, because now it is indeed 1 map parameter.
I'll rollback changes

eugenemedvediev: Done

ckipp01: Yea, pros and cons here basically. I miss being able to use the literal call syntax, but I also like opinionated auto formatting. So that wins for me here. There was actually a discussion about this a while back if you're curious https://github.com/JohnnyMorganz/StyLua/issues/65

ckipp01 (approved):
👍🏼 thanks for taking care of this @eugenemedvediev, and I appreciate the detailed pr description.

ckipp01:
@mergifyio backport next

PR #180, opened 2021-07-14
Align Metals neovim commands with Metals Telescope commands

eugenemedvediev:
I was curious about neovim commands misalignment you mentioned on Twitch, so I decided to experiment.
Everything seems to work and looks now more mature.
What do you think about this change?
Changes:
- Fixed typo `luacheck`
- Fixed typo for `MetalLogToggle` ( missing `s` after Log ) ( changed later to `MetalsToggleLogs` )
- Fixed typo for `MetalsDoctor` in documentation and changed to `MetalsRunDoctor`
- Changed neovim Metals command in the way where the verb is followed by a noun
- Reflected commands changes in documentation
- checked format and lint
- Sorted metals telescope commands alphabetically
- Fixed `frop` typo in `analyze_stacktrace`
- Renamed `worksheet_hover` to `hover_worksheet` command
- Sorted Commands in docs
- Sorted LUA API commands in docs
- Added support for previous versions of commands ( errors for commands, for which it is impossible to show a warning, as it disappears; for example for `logs_toggle`. I've added warnings for commands which allow the message to be displayed while new command is being executed)

### DIFF
--- a/doc/metals.txt
+++ b/doc/metals.txt
@@ -415,6 +415,7 @@ The following commands are provided by nvim-metals:
   * |MetalsSwitchBsp|
   * |MetalsToggleLogs|
   * |MetalsUpdate|
+
                                                        *MetalsAnalyzeStacktrace*
 MetalsAnalyzeStacktrace      Takes the stacktrace from your current register
                              and sends it to Metals to be analyzed. This will
--- a/doc/metals.txt
+++ b/doc/metals.txt
@@ -388,39 +388,39 @@ handle it.
 COMMANDS                                                       *metals-commands*
 
 The following commands are provided by nvim-metals:
-  * |MetalsAmmoniteStart|
-  * |MetalsAmmoniteStop|
+  * |MetalsStartAmmonite|
+  * |MetalsStopAmmonite|
   * |MetalsAnalyzeStacktrace|
-  * |MetalsBspSwitch|
-  * |MetalsBuildConnect|
-  * |MetalsBuildDisconnect|
-  * |MetalsBuildImport|
-  * |MetalsBuildRestart|
+  * |MetalsSwitchBsp|
+  * |MetalsConnectBuild|
+  * |MetalsDisconnectBuild|
+  * |MetalsImportBuild|
+  * |MetalsRestartBuild|
   * |MetalsCompileCancel|
   * |MetalsCompileClean|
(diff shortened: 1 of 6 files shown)


doc/metals.txt
```diff
@@ -388,39 +388,39 @@ handle it.
 COMMANDS                                                       *metals-commands*
 
 The following commands are provided by nvim-metals:
-  * |MetalsAmmoniteStart|
-  * |MetalsAmmoniteStop|
+  * |MetalsStartAmmonite|
```
ckipp01, line 391: Sort of a nitpick, but if we are going to change this all around, can we keep them alphabetical and rearrange them?

lua/metals.lua
```diff
@@ -34,31 +34,31 @@ M.analyze_stacktrace = function()
   end
 end
 
-M.ammonite_start = function()
+M.start_ammonite = function()
```
ckipp01, line 37: So if we change these, we'd need to also change them under the Lua API section in the docs.

I like this, but also realize that this will break any mappings people may have that are using these. Let's put a note in the pr description that notifies people that this _does_ technically break the Lua API `nvim-metals` provides and also a mapping of `old -> new` to so it's crystal clear for someone.

ckipp01 (changes requested):
Thanks for the pr @eugenemedvediev! Couple nits and suggestions.

doc/metals.txt
eugenemedvediev: Done

lua/metals.lua
eugenemedvediev: Documentation is changed.
I need to think about how correctly inform about changed commands.
I think fugitive has a nice way to inform about deprecated mappings.

eugenemedvediev: I've added deprecated and not supported variants for the commands.

From tomorrow till 28 June I am on vacation, so, unfortunately, I can't make any changes to this PR before 28.
So if this solution is not mature enough, then I'll pick it in 2 weeks. :)

ckipp01 (approved):
While it's a bit annoying to have both for a bit, I think this is the best choice. We'll leave them in for a month or something and then rip them out. 👍🏼 thanks a lot for taking care of this.

ckipp01:
@mergifyio backport next

PR #191, opened 2021-07-29
Add in ability to toggle boolean settings.

### LINKED ISSUE #189: Possibility to toggle showing implicits
### Discussed in https://github.com/scalameta/nvim-metals/discussions/188

<div type='discussions-op-text'>

<sup>Originally posted by **stumash** July 22, 2021</sup>
I see that there are configuration options on setup for showing `implicit` arguments and conversions.

Is there a way to toggle these options _while `metals` is running_?</div>

ckipp01:
This will allow for a user to easily use a mapping to toggle a boolean
setting. Thinking about this a bit more, we can probably in the future
make a nice way to change any setting value and send that in. Something
that prompts the user for input, and then they can input the new value
and the same logic could be used to send in the notification to Metals.
But until we need that, this should be fine.

Closes #189

Pinging @stumash since you were the one to bring this up.

### DIFF
--- a/doc/metals.txt
+++ b/doc/metals.txt
@@ -657,9 +657,10 @@ initialize_or_attach({config})
 
   api.nvim_command [[augroup NvimMetals]]
   api.nvim_command [[autocmd!]]
-  api.nvim_command [[autocmd BufEnter <buffer> lua require("metals").did_focus()]]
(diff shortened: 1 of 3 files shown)


stumash:
awesome 🙌

PR #198, opened 2021-08-12
Fix bug where metals.lua doesn't see config changes

### LINKED ISSUE #197: `:MetalsRestartServer` fails with error `attempt to index local 'config' (a nil value)`
**Describe the bug**
Running `:MetalsRestartServer` results in the error
```
Error executing vim.schedule lua callback: /Users/cgraff/.vim/plugged/nvim-metals/lua/metals/setup.lua:228: attempt to index local 'config' (a nil value)                                                                     
```

The bug was caused by 474151a9120f0939dc71a40abf58ff86e17336a3.

By changing from the `M = {}` export semantic to individually exporting local functions, now when `metals.lua` calls `setup.config_cache` the value is always nil, because `setup.lua` exports `local config_cache` and `metals.lua` does not see mutations to the `config_cache` made by `setup.initialize_or_attach`

camgraff:
Fixes #197

### DIFF
--- a/lua/metals.lua
+++ b/lua/metals.lua
@@ -313,7 +313,7 @@ M.restart_server = function()
 
   vim.defer_fn(function()
     setup.reset_lsps()
-    setup.initialize_or_attach(setup.config_cache)
+    setup.initialize_or_attach()
   end, 3000)
 end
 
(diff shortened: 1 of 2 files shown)


ckipp01 (approved):
Thanks a bunch for catching this @camgraff! And thanks for sending in a fix!

PR #199, opened 2021-08-14
Fix autocmd code example in help

greenhat:


ckipp01 (approved):
Ah yes, oversight on my part. I can see how this could be confusing as well if I don't explain what `metals_config` is before this. Thanks for the fix!

PR #213, opened 2021-09-07
Set superMethodLensesEnabled to default to on.

ckipp01:
In the last release of metals there were some massive improvements
to the speed of this feature. This used to be turned off due to
performance issues, which we no longer need to worry about.


ckipp01:
@jsatk btw you may want to do this in `coc-metals` as well. We do it in VS Code too.

jsatk:
> @jsatk btw you may want to do this in `coc-metals` as well. We do it in VS Code too.

Will get to it today or tomorrow. Thanks for the heads up.

PR #218, opened 2021-09-14
Fix handler compat for requests.

### LINKED ISSUE #217: MetalsNewFile: error in method "metals/quickPick"
**Describe the bug**

Attempting to create a new worksheet, in nvim I run `:MetalsNewScalaFile` and enter `6` in the prompt (this is the Worksheet option).
I get an error reported:
```
Error executing vim.schedule lua callback: ...nt_nvimHlwrLf/usr/share/nvim/runtime/lua/vim/lsp/rpc.lua:511: method
"metals/quickPick": either a result or an error must be sent to the server in response
```

1. Open a file in my current test project
2. Go to `build.sbt`
3. Ensure the build is imported
4. DO this `:MetalsNewScalaFile`
5. pick any option
6. See error

**Expected behavior**

A new worksheet is created.

**Installation:**

- Operating system: Linux
- Nvim version:  NVIM v0.6.0-dev+223-g384a43fe7
- Commit of nvim-metals: 954c6d4
- Metals version: metals 0.10.6

**Search terms**

MetalsNewScalaFile

ckipp01:
I don't fully understand why the compat function won't work
as expected for requests that expect something back. So this
fixes the issue, but I feel like it shouldn't be necessary. I'll
continue to look into this to see if there is somethign else I can do
to better align this, but for now this will fix it.

Close #217

### DIFF
--- a/lua/metals/handlers.lua
+++ b/lua/metals/handlers.lua
@@ -14,7 +14,9 @@ local decoration_namespace = api.nvim_create_namespace("metals_decoration")
 
 -- Implementation of the `metals/quickPick` Metals LSP extension.
 -- - https://scalameta.org/metals/docs/integrations/new-editor.html#metalsquickpick
-M["metals/quickPick"] = util.lsp_handler(function(_, result)
+M["metals/quickPick"] = function(...)
+  local result = util.extract_handler_result(...)
+
   local ids = {}
   local labels = {}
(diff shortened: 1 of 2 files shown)


ckipp01:
Use https://github.com/scalameta/nvim-metals/pull/220 instead

PR #221, opened 2021-09-15
Fix display of diagnostic signs

kluen:
The naming layout for the diagnostic signs has changed:
https://github.com/neovim/neovim/pull/12655

### DIFF
--- a/plugin/metals.vim
+++ b/plugin/metals.vim
@@ -8,7 +8,7 @@ endfunction
 
 function! metals#errors() abort
   let errorCount = luaeval("vim.lsp.diagnostic.get_count(vim.fn.bufnr('%'), [[Error]])")
-  let possibleLspSign = sign_getdefined("LspDiagnosticsErrorSign")
+  let possibleLspSign = sign_getdefined("LspDiagnosticsSignError")
   let sign = get(possibleLspSign, 0, {"text": "E"})
   if (errorCount > 0)
     return sign.text . errorCount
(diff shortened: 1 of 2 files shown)


plugin/metals.vim
```diff
@@ -8,7 +8,7 @@ endfunction
 
 function! metals#errors() abort
   let errorCount = luaeval("vim.lsp.diagnostic.get_count(vim.fn.bufnr('%'), [[Error]])")
-  let possibleLspSign = sign_getdefined("LspDiagnosticsErrorSign")
+  let possibleLspSign = sign_getdefined("LspDiagnosticsSignError")
```
ckipp01, line 11: So I'd like to still ensure that this works for 0.5 users, and if I understand this correctly, this will only correctly work for those that are using nightly. What do you think about trying to detect 0.5 or nightly and correctly grabbing the right one to get the count?

kluen: Hi Chris, sounds like a good idea. However, I just checked, the change is from Nov 2020.
Looks like it is in 0.5 already: https://github.com/neovim/neovim/blob/release-0.5/runtime/lua/vim/lsp/diagnostic.lua#L631-L634
Or am I misunderstanding anything? I am still a bit new to the internals of neovim and nvim-metals. 😅

ckipp01: Ahh, that's totally my mistake! I quickly looked at the date of the last "merged" icon and saw it was ten days ago not realizing it was a linked pr. Thanks for catching this!

ckipp01 (approved):

PR #226, opened 2021-09-21
Javap semanticdb command support

ckipp01:
- [x] The two `javap` commands aren't working.  It always returns `Error: class not found: Thing.class`

NOTE: The above is still an issue, but only in the situation where the package structure on disk doesn't match the package structure you actually have in place.
- https://github.com/scalameta/metals/issues/3150
- https://github.com/scalameta/metals/issues/3136

### DIFF
--- a/lua/metals.lua
+++ b/lua/metals.lua
@@ -2,6 +2,7 @@ local api = vim.api
 local fn = vim.fn
 local lsp = vim.lsp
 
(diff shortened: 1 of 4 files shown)


ckipp01:
Also adds in the changes necessary for https://github.com/scalameta/metals/pull/3148

PR #229, opened 2021-10-01
Refactor out the UI module with Plenary

ckipp01:
Note that this does require users to have plenary installed. I've
hesitated to do this for quite some time, but I figured most people are
probably using this anyways for other stuff, so why not simplify my
life.


ckipp01:
Closes #125

PR #232, opened 2021-10-03
Add suport for type on code selection.

### LINKED ISSUE #228: Add support for type annotation on code selection
**Task**
Now that https://github.com/scalameta/metals/pull/3060 is merged in we'll need to add support for this. We probably need to tweak a few things and do some sort of custom binding since this expects a range in, not just a point like the normal hover.

ckipp01:
This adds in initial support, but this issue in Metals
should be merged in first: https://github.com/scalameta/metals/pull/3178.

Closes #228


ckipp01:
I'm going to pause on this as I'm struggling with a good way to actually get this to work. Wait for an answer in https://github.com/neovim/neovim/issues/15905

PR #237, opened 2021-10-05
Refactor decode file.

ckipp01:
These changes are necessary for the changes coming in
https://github.com/scalameta/metals/pull/3160.


lua/metals.lua
```diff
@@ -276,84 +276,61 @@ M.restart_server = function()
...
-M.show_tasty = function()
-  -- TODO don't send in URI, send in textDocumentPositionParams
-  local text_doc_position = lsp.util.make_position_params()
-  execute_command({ command = "metals.show-tasty", arguments = { text_doc_position } })
-end
-
-M.start_server = function()
-  if vim.g.metals_disabled_mode then
-    setup.explicitly_enable()
+local function show_decoded(decoder_type, format)
```
ckipp01, line 279: Should maybe add a check in here to not allow a user to use this if there are unsaved changes in the buffer.

PR #246, opened 2021-10-08
Refactor configuration building out of setup and start testing.

ckipp01:
This PR breaks out the configuration building into a new `configuration.lua` which handles validating a configuration before we try passing it in to core. This cleans up `setup.lua` a bit, but more importantly allows us to start unit testing chunks of `nvim-metals`. So this also adds in a base framework using the test harness from Plenary to start checking the configuration build up.


ckipp01:
I seem to be hitting on https://github.com/neovim/neovim/issues/15709 which is causing the tests to not run. Just keep an eye on this and hopefully it gets resolved before this is ready to merge.

PR #247, opened 2021-10-08
Basic `findTextInDependencyJars` implemented.

ckipp01:
This is needed for the functionality added in
https://github.com/scalameta/metals/pull/3093.


ckipp01:
For now I just stole the logic from https://github.com/neovim/neovim/pull/15960, however if that gets merged we should be able to just use the logic straight from core.

PR #258, opened 2021-10-18
Add option to override find_root_dir function

### LINKED ISSUE #257: Support even more multi-module project hierarchies
**Describe the feature**

In #6 (and then subsequently in #37), the [`find_root_dir`](https://github.com/scalameta/nvim-metals/blob/0d5ff68afcd2db4bb35acd516229dec80092f176/lua/metals/util.lua#L23) function was added to support nested project roots. The [`validate_config` function](https://github.com/scalameta/nvim-metals/blob/0d5ff68afcd2db4bb35acd516229dec80092f176/lua/metals/config.lua#L131) uses that function to set the `root_dir` LSP config string. I would love to override that setting to support some stranger multi-module project hierarchies (see "Additional context").

I don't know exactly how we want to make that happen, but I currently have two ideas:

1. Change that line of `validate_config` to allow users to set their own root dir, like this:
```lua
config.root_dir = config.root_dir or util.find_root_dir...
```
2. Add more complicated logic to the `find_root_dir` function to support weirder nestings of project hierarchies.

I would imagine that most users are happy with the existing setup, and going with option 2 might lead to inadvertent problems for those users. Going with option 1 would be pretty simple, and allow users to customize this however they want. If this strange project hierarchy (see "Additional context") ends up being a common use-case (I'd be surprised if that happened, but who knows...), then we could revisit option 2 in the future.

**Additional context**

I regularly work on a project that has a directory structure that looks like this:

```
./
├── build.sbt
├── dir1/
│  └── dir2/
│     ├── build.sbt
│     └── src/
│        └── main/
│           └── scala/
└── src/
   └── main/
      └── scala/
```

Right now, if I ever open a `.scala` file underneath the `dir2` hirerarchy, the root-level `build.sbt` file won't be found. This is due to how the [`find_root_dir`](https://github.com/scalameta/nvim-metals/blob/0d5ff68afcd2db4bb35acd516229dec80092f176/lua/metals/util.lua#L23) function is implemented, which looks for root files either in the current directory or the direct parent. This organization would require looking yet another level above.

**Search terms**

config, root_dir

tymcauley:
This new function is part of the config table passed into
`initialize_or_attach()`. For example:

```lua
metals_config = require("metals").bare_config()
metals_config.find_root_dir = function(patterns, startpath)
  -- insert your functionality here...
end
```

The `patterns` argument is `metals_config.root_patterns`, and the
`startpath` argument is the path to the current buffer. The return value
of this function sets the `root_dir` parameter of the LSP client
configuration table. See `:help vim.lsp.start_client` for more
information.

Resolves #257.

### DIFF
--- a/lua/metals/config.lua
+++ b/lua/metals/config.lua
@@ -128,7 +128,9 @@ local function validate_config(config, bufnr)
 
   local bufname = api.nvim_buf_get_name(bufnr)
 
-  config.root_dir = util.find_root_dir(config.root_patterns, bufname) or fn.expand("%:p:h")
+  local find_root_dir = config.find_root_dir or util.find_root_dir
+
+  config.root_dir = find_root_dir(config.root_patterns, bufname) or fn.expand("%:p:h")
 
   local base_handlers = vim.tbl_extend("error", default_handlers, tvp.handlers)
 


tymcauley:
At this point, the documentation is a bit out-of-date, since it references `root_patterns` as the only special key passed into `initialize_or_attach()`. Want me to take a crack at updating the docs?

PR #269, opened 2021-11-07
Add test for detecting root dir when non root

ckipp01:
Interesting, this fails when you give it both `.scala` and `.git`, but not if you give it `.git`. Something odd with the ordering is going on here I assume.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 doc/tags
 mill-minimal/
 multiple-build-file-example/
+minimal-scala-cli-test/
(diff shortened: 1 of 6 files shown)


ckipp01:
@mpasa I'm not sure this is even the issue you were hitting on, but an issue nonetheless.

mpasa:
@ckipp01 In my case, it doesn't work either having only `.git` or `.scala` and `.git`. I will play with the `rootdir.lua` file to see if I can get some insights.

mpasa:
@ckipp01 Well, forget about my previous comment. I incorrectly updated the plugin. The main branch after this merge works as expected, with both options. Thank you!

ckipp01:
> @ckipp01 Well, forget about my previous comment. I incorrectly updated the plugin. The main branch after this merge works as expected, with both options. Thank you!

Great, glad to hear it works!

PR #270, opened 2021-11-07
corrected typo

softinio:


ckipp01 (approved):
Thanks for catching these @softinio and sending in a pr!

PR #271, opened 2021-11-09
Correct packer.nvim syntax

ayhon:
In lua there are no named arguments, this is simulated by the fact that when passing a table or string literal, the parenthesis can be omitted. `use` expects a table with its first value being the direction of the extension, and among other optional fields, a table in the key `requires` with the directions of the extensions required. 

In the notes, it shows as if `use` takes two arguments, a table with the direction of the first extension and then a named argument with a table with the direction of the first extension.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ it.
 example if using [`packer`](https://github.com/wbthomason/packer.nvim):
 
 ```lua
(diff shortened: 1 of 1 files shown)


ayhon:
As it stands, it just fails

ckipp01 (approved):
Thanks a lot for catching this @ayhon. This wasn't intentional and I think a leftover remnant of  before I had a reliance on plenary.

PR #297, opened 2021-12-12
Fix Windows path, support .bat files

octavz:
I propose a small fix for now, it will check if both metals.bat and cs.bat are on the path and uses them. It should work.
Both metals and coursier execs should be configurable and have different default values on win/linux.

### DIFF
--- a/lua/metals/config.lua
+++ b/lua/metals/config.lua
@@ -39,7 +39,7 @@ end
 
 --- Used to generate Metals installation file name depending on configuration
 --- @return string fileame
-local function metals_filename()
+local function metals_install_name()
   if config_cache and config_cache.settings.metals.useGlobalExecutable then
     return metals_name
   else
@@ -55,10 +55,9 @@ end
 local function metals_bin()
   local metals_filename = metals_name
   -- Coursier will install both a "metals" and "metals.bat", on Windows we should use "metals.bat" as exe
-  local metals_win_filename = metals_name .. ".bat"
 
-  if util.has_bins(metals_win_filename) then
-    metals_filename = metals_win_filename
+  if util.is_windows then
+    metals_filename = metals_name .. ".bat"
   end
 
   if config_cache and config_cache.settings.metals.useGlobalExecutable then
@@ -70,15 +69,17 @@ end
 
 --- Check to see if coursier is installed. This method favors the native cs. So if
 --- cs is installed, that will be returned, if not, then coursier will be returned.
---- @return string 'cs', 'coursier', or nil
+--- @return string 'cs', 'cs.bat', 'coursier', or nil
 local function check_for_coursier()
   -- this must be first the second "cs" is also found as binary but we should be more specific on Windows
-  if util.has_bins("cs.bat") then
+  if util.is_windows then
     return "cs.bat"
-  elseif util.has_bins("cs") then
-    return "cs"
-  elseif util.has_bins("coursier") then
-    return "coursier"
+  else
+    if util.has_bins("cs") then
+      return "cs"
+    elseif util.has_bins("coursier") then
+      return "coursier"
+    end
   end
 end
 
@@ -297,7 +298,7 @@ return {
   in_disabled_mode = in_disabled_mode,
   get_config_cache = get_config_cache,
   metals_bin = metals_bin,
-  metals_filename = metals_filename,
+  metals_install_name = metals_install_name,
   metals_init_options = metals_init_options,
   scala_file_types = scala_file_types,
   set_config_cache = set_config_cache,
--- a/lua/metals/config.lua
+++ b/lua/metals/config.lua
@@ -37,12 +37,9 @@ local function in_disabled_mode(config)
   return disabled
 end
 
---- Ultimately what will be passed to the config.cmd to initialize the LSP
---- connection. If a user is using useGlobalExecutable then we just default
---- to `metals`, if not we take control and construct the location in the
---- cache dir.
---- @return string executable
-local function metals_bin()
+--- Used to generate Metals installation file name depending on configuration
+--- @return string fileame
+local function metals_filename()
   if config_cache and config_cache.settings.metals.useGlobalExecutable then
     return metals_name
   else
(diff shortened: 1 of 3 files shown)


lua/metals/config.lua
```diff
@@ -37,24 +37,45 @@ local function in_disabled_mode(config)
   return disabled
 end
 
+--- Used to generate Metals installation file name depending on configuration
+--- @return string fileame
+local function metals_filename()
```
ckipp01, line 42:
```suggestion
local function metals_executable_name()
```
If now we'll get warnings about a redefined local down below where we also are defined a `metals_filename`. So we can either change it there or change it here, but let's make them different.

lua/metals/config.lua
```diff
@@ -37,24 +37,45 @@ local function in_disabled_mode(config)
...
+end
+
 --- Ultimately what will be passed to the config.cmd to initialize the LSP
 --- connection. If a user is using useGlobalExecutable then we just default
 --- to `metals`, if not we take control and construct the location in the
 --- cache dir.
 --- @return string executable
 local function metals_bin()
+  local metals_filename = metals_name
+  -- Coursier will install both a "metals" and "metals.bat", on Windows we should use "metals.bat" as exe
```
ckipp01, line 57: Out of curiosity, why does it install a `metals` on windows then if it needs to use `metals.bat`?

lua/metals/config.lua
ckipp01: This function ends up pretty much returning exactly what `metals_bin()` returns, but I see you use this now in the install instead of `metals_bin()`. Aren't we able to just re-use that?

lua/metals/config.lua
```diff
@@ -37,24 +37,45 @@ local function in_disabled_mode(config)
...
   end
 end
 
 --- Check to see if coursier is installed. This method favors the native cs. So if
 --- cs is installed, that will be returned, if not, then coursier will be returned.
 --- @return string 'cs', 'coursier', or nil
 local function check_for_coursier()
-  if util.has_bins("cs") then
+  -- this must be first the second "cs" is also found as binary but we should be more specific on Windows
+  if util.has_bins("cs.bat") then
```
ckipp01, line 76: So there is a few different places that we are checking for `cs.bat` or `metals.bat` even when we aren't on windows. Let's not do this, but instead check the os, and only look for the bat versions if we are on window. You can do this with `vim.loop.os_uname()` which returns a table of info that you can look at. You'll probably want to look at `sysname` in there.

ckipp01 (changes requested):
Thanks for the pr @octavz! I left a few comments because I think we can simplify this a bit and I also have a couple questions.

lua/metals/config.lua
octavz: metals_bin returns the actual binary, this function returns the installation name for metals, perhaps name can be improved `metals_install_name` or something, but they are not the same on windows.

lua/metals/config.lua
octavz: I think is a linux exe, or equivalent, has some bash in the header and the rest is binary

lua/metals/config.lua
octavz: a bit complicated given my [non existent] Lua skills but will try.

octavz:
@ckipp01 please have another look, I used `packer`, `is_windows` function, probably way better tested than what I can write.

lua/metals/config.lua
ckipp01: Ahh cause even on windows you want to install metals, but you end up with `metals.bat`. Got it.

lua/metals/config.lua
ckipp01: No worries. I do it a different way in the `.jvmopts` file, but I'll extract that into a utility function and change it later.

ckipp01:
Great, thanks @octavz. I'd like to have another couple windows users verify that this works before merging in.

octavz:
can you help with the tests, how do I run them ?

ckipp01:
> can you help with the tests, how do I run them ?

I think the test failure may be flaky. I re-triggerd it. You can see instructions on testing here https://github.com/scalameta/nvim-metals/blob/main/CONTRIBUTING.md#testing

octavz:
> > can you help with the tests, how do I run them ?
> 
> I think the test failure may be flaky. I re-triggerd it. You can see instructions on testing here https://github.com/scalameta/nvim-metals/blob/main/CONTRIBUTING.md#testing

seems fine now . Please let me know if I have to do anything, I don't follow the discord channel as much as you.

zetashift:
Tested the branch on Windows 10, everything worked fine!

image

ckipp01:
Thanks for testing @zetashift and thanks again for the contribution @octavz! Hopefully this will help windows users in the future.

lcharles:
Hi, I just installed nvim-metals on Windows and here some feedback on the installation process.
I think that posting it here is more relevant then creating a new issue.
Following the doc on Coursier's website I installed the native version which is a cs.exe. With this setup MetalsInstall will fail as it only looks for cs.bat. Changing cs.bat to cs.exe in config.lua:78 allow MetalsInstall to run but it should probably probe for both to allow both CS installation methods. Otherwise maybe specify in the doc what installation is required.

Cheers,
If there is anything you want me to test on windows just let me know.

ckipp01:
Thanks a lot for the feedback @lcharles. @octavz, what installation method did you use that you ended up with a `cs.bat`?

>  With this setup MetalsInstall will fail as it only looks for cs.bat. Changing cs.bat to cs.exe in config.lua:78 allow MetalsInstall to run but it should probably probe for both to allow both CS installation methods.

Maybe I can just do this then and if on windows check for both? Although I am a bit confused, why there should even need to be two after looking at the docs. I'll wait to see about how @octavz got his `cs.bat`.

ckipp01:
Actually looking at this a bit more, I guess the `.bat` comes from the jar launcher? https://get-coursier.io/docs/cli-installation#windows-1
So is it safe to say that if you have cs installed as a native launcher on windows the `cs.exe` _should_ be able to be just triggered with `cs` no?

octavz:
@ckipp01 I don't remember but I doubt I did something else different from what they recommended at that time.
Ended up with this, no `.exe` or `.jar` file in the path, honestly I don't even know how it works, the `bat` files have some binary stuff at the end that gets executed.

image

octavz:
I think I ran this:
https://github.com/coursier/launchers/raw/master/cs-x86_64-pc-win32.exe 

Running this will generate `bat` files in that folder.

lcharles:
> Actually looking at this a bit more, I guess the `.bat` comes from the jar launcher? https://get-coursier.io/docs/cli-installation#windows-1 

Yes it seems to be the case. It also comes from running cs.exe. 
In fact I think I just tricked myself. 
> So is it safe to say that if you have cs installed as a native launcher on windows the `cs.exe` _should_ be able to be just triggered with `cs` no?

In fact cs.exe seems to only be an installer that you run once to setup the appdata folder. And once you add it to your path you can use it as in `cs install metals` from the shell (this will call cs.bat). Apparently It just happens that it can also be called to install stuff as well.

It now seems to me that this problem only happened because I thought that putting cs.exe anywhere in my path would work. And obviously since it is explicitly looking for cs.bat it didn't. 
Whereas I had to run it once and add appdata/coursier/data/bin to my path. In my defense coursier's doc is not super precise about that.

Now to avoid this to happen again, as it should be really rare, in my opinion it is enough to inform users that nvim-metals expects that the users has run cs.exe once and that he has appdata/coursier/data/bin in his path and suggesting to run cs --help from a new terminal session to verify it. Maybe also adding more information in the error message if it fails would be enough.

ckipp01:
> Now to avoid this to happen again, as it should be really rare, in my opinion it is enough to inform users that nvim-metals expects that the users has run cs.exe once and that he has appdata/coursier/data/bin in his path and suggesting to run cs --help from a new terminal session to verify it. Maybe also adding more information in the error message if it fails would be enough.

Sure, are you able to send in a pr? We can put something like this with a special note for windows users under the prerequisites section where it mentions coursier.

lcharles:
> > Now to avoid this to happen again, as it should be really rare, in my opinion it is enough to inform users that nvim-metals expects that the users has run cs.exe once and that he has appdata/coursier/data/bin in his path and suggesting to run cs --help from a new terminal session to verify it. Maybe also adding more information in the error message if it fails would be enough.
> 
> Sure, are you able to send in a pr? We can put something like this with a special note for windows users under the prerequisites section where it mentions coursier.

Sure, I'll do that during the day.

PR #314, opened 2021-12-31
Improves Coursier setup documentation and error reporting on Windows.

lcharles:
Hi here is my quick fix for the problem I reported 2 days ago in https://github.com/scalameta/nvim-metals/pull/297.
I could not see the backslashes in the path in the error message so I left it with normal slashes.

### DIFF
--- a/lua/metals/config.lua
+++ b/lua/metals/config.lua
@@ -75,7 +75,13 @@ end
 local function check_for_coursier()
   -- this must be first the second "cs" is also found as binary but we should be more specific on Windows
   if util.is_windows then
-    return "cs.bat"
+    if util.has_bins("cs.bat") then
+      return "cs.bat"
+    else
+      log.error_and_show("Warning:Could not find cs.bat in PATH")
+      log.error_and_show("Installation instructions for Coursier https://get-coursier.io/docs/cli-installation#windows")
+      log.error_and_show("Make sure to run cs.exe once to install Coursier and add C:/Users/YOURNAME/AppData/Coursier/data/bin to your path.")
+    end
   else
     if util.has_bins("cs") then
       return "cs"
--- a/lua/metals/config.lua
+++ b/lua/metals/config.lua
@@ -78,9 +78,7 @@ local function check_for_coursier()
     if util.has_bins("cs.bat") then
       return "cs.bat"
     else
-      log.error_and_show("Warning:Could not find cs.bat in PATH")
-      log.error_and_show("Installation instructions for Coursier https://get-coursier.io/docs/cli-installation#windows")
-      log.error_and_show("Make sure to run cs.exe once to install Coursier and add C:/Users/YOURNAME/AppData/Coursier/data/bin to your path.")
+      log.error_and_show(messages.coursier_not_installed_windows)
     end
   else
     if util.has_bins("cs") then
--- a/README.md
+++ b/README.md
@@ -29,6 +29,10 @@ integrations with other projects such as
     newer.
 - Ensure [Coursier](https://get-coursier.io/docs/cli-installation) is installed
     on your machine. `nvim-metals` uses Coursier to download and update Metals.
+    _NOTE_: On Windows, you should run `coursier` or `cs.exe` once from the command
+    line as this is how it will install itself. Once this is done you should 
+    add `C:\Users\YOURNAME\AppData\Coursier\data\bin` to your path. To verify
+    that it is properly installed you can run `cs --help` from a new shell.
 - Remove `F` from `shortmess`. `set shortmess-=F`
     (for lua `vim.opt_global.shortmess:remove("F")`)
     _NOTE_: Without doing this, autocommands that deal with filetypes prohibit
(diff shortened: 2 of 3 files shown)


lua/metals/config.lua
```diff
@@ -75,7 +75,13 @@ end
 local function check_for_coursier()
   -- this must be first the second "cs" is also found as binary but we should be more specific on Windows
   if util.is_windows then
-    return "cs.bat"
+    if util.has_bins("cs.bat") then
+      return "cs.bat"
+    else
+      log.error_and_show("Warning:Could not find cs.bat in PATH")
```
ckipp01, line 81: Let's combine these three into one call, and put it in `messages.lua`.  You can see some other example there. I believe to see the `/` you just need to escape it. Can you try with `\\`?

ckipp01 (changes requested):
Thanks for sending this in! Just a quick nit.

lua/metals/config.lua
lcharles: > Let's combine these three into one call, and put it in `messages.lua`. You can see some other example there.

Ok, I see that you already have a message for when coursier is not installed. 
```M.coursier_not_installed = [[
It looks like you don't have Coursier installed, which you need to install Metals.

You can find instructions on how to install it here: https://get-coursier.io/docs/cli-installation]]
```
Do you want me to use this one or to create another one more specific for Windows. I think another one would be better since the instructions on coursier's website does not mention to add coursier to the path. I could also modify this one instead to be more specific for windows.

> I believe to see the `/` you just need to escape it. Can you try with `\\`?

I tried it first like this `C:\\Users\\YOURNAME\\AppData\\Coursier\\data\\bin` but the backslashes would not appear in the final string, it gave me the string  `C:UsersMynameAppDataCoursierdata^Hin `.

ckipp01: I'd say having a windows specific message here may be best. So we can detect the OS and then show then the windows one and for everything else we an use the existing one.

As for the path, that's really odd. I'm not sure why that's happening. I'd honestly rather keep that part out then potentially confuse people by putting `/path` format with windows.

lcharles: > I'd say having a windows specific message here may be best. So we can detect the OS and then show then the windows one and for everything else we an use the existing one.
> 
Ok, here's a proposal:
```
Warning: Could not find cs.bat in PATH
You can find instructions on how to install it here: https://get-coursier.io/docs/cli-installation#windows
Make sure to run cs.exe once to install Coursier and add it's installation directory to your path. 
See the note in the Readme.
```
> As for the path, that's really odd. I'm not sure why that's happening. 

Yes me neither. I tried `:lua print("C:\\Users\\YOURNAME\\AppData\\Coursier\\data\\bin")` and it escapes everything properly. 

>I'd honestly rather keep that part out then potentially confuse people by putting `/path` format with windows.

Ok, I'll just make a reference to the Readme.

lua/metals/config.lua
```diff
@@ -75,7 +75,13 @@ end
...
+      log.error_and_show("Installation instructions for Coursier https://get-coursier.io/docs/cli-installation#windows")
+      log.error_and_show("Make sure to run cs.exe once to install Coursier and add C:/Users/YOURNAME/AppData/Coursier/data/bin to your path.")
```
ckipp01, line 83: Sorry if I wasn't clear in the last comment, but let's remove this line. Since it won't show correctly it's better not to have it. Also you'll need to ran `make format` to ensure everything is formatted correctly.

lcharles: > Sorry if I wasn't clear in the last comment, but let's remove this line. Since it won't show correctly it's better not to have it. 

Now, I am confused ;).
I planned to replace the three log calls with a single one and put the text in `messages.lua`. And the message would be what I suggest above. Is it ok ? I suggested the text here before commiting it to avoid polluting the logs.

> Also you'll need to ran `make format` to ensure everything is formatted correctly.

I installed stylua and ran it but I don't know if it works because it does not give me any output.

ckipp01: Ach! Sorry, lol that was my mistake. Yes, your proposed text was fine. I totally misread the whole thing 😬 .

lcharles: Haha no worries. I'll do another pr right away.
BTW I am wondering how do you manage the dev and usage of nvim.metals.
Say you want to use it regularly but also want to hack on it from time to time. For example in my case I installed scalameta/metals, but to hack on it I need to pull and push to my github account. As I don't really know how packer manages this I did not swap it yet with my github account.

ckipp01: Good question. What I'd recommend is just cloning this repo down on your local machine where you want to work on it, and then instead of installing `ckipp01/nvim-metals`, just install it from that local place on your machine. You can see how I do this with packer in my dot files [here](https://github.com/ckipp01/dots/blob/35ef993338165a2ea7d22a508415eb11eae63cb4/nvim/.config/nvim/lua/mesopotamia/plugins/init.lua#L32-L35)

lcharles: Great thx, I just did a pull request, I just had an issue with make format, because it would change all line endings. I was able to catch it but it might be good to add a rule for this in the Makefile. I'll look into it later.

ckipp01 (approved):
Thanks for this @lcharles! I'm create another pr shortly on top of this because there is one other thing I'd like to change. In the future, if possibly create your pr not on your main branch as it then prohibits me from pushing on top of it.

LGTM!

PR #325, opened 2022-01-18
Explicitly pass character offset encoding to `apply_workspace_edit`

suliatis:
this fixes the `MetalsOrganizeImport` command in neovim nightly.
  
  Here is the corresponding breaking change announcement: https://github.com/neovim/neovim/issues/14090#issuecomment-1012005684

### DIFF
--- a/lua/metals.lua
+++ b/lua/metals.lua
@@ -311,7 +311,7 @@ M.organize_imports = function()
   for _, response in pairs(resp) do
     for _, result in pairs(response.result or {}) do
       if result.edit then
-        lsp.util.apply_workspace_edit(result.edit, 'utf-16')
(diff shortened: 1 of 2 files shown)


ckipp01:
Thanks for the pr @suliatis! There is actually one more needed one here: https://github.com/scalameta/nvim-metals/blob/d5864d9395152d4820bcb8a66f1235c5970e46a5/lua/metals/handlers.lua#L47 Can you also update that one?

suliatis:
Sure. Unfortunately I'm not sure about the impact of this other change, though should not break anything.

ckipp01 (approved):
Looks good, thanks for the contribution!

PR #348, opened 2022-03-15
feat: take into account tooltip and command in metals/status

### LINKED ISSUE #102: Handle `metals/status` command instead of just the text.
**Describe the feature**

So both in this extension and in coc-metals I've sort of implemented _half complete_ `metals/status` to benefit from the much better messages. However, that's only part of what `metals/status` can actually do. Take this example of when you try to run a program that has errors in the workspace:

```
[Trace - 09:21:01 AM] Sending notification 'metals/status'
Params: {
  "text": "$(error) Errors in workspace",
  "tooltip": "Cannot run or debug due to existing errors in the workspace. Please fix the errors and retry.",
  "command": "metals-diagnostics-focus"
}
```

There is a lot of extra information in here, but we only capture the `text`. Think of a way to be able to capture the `tooltip` and `command`.

I'm not 100% sure how to display this in an unobtrusive way. Maybe with a popup window of some sort, or just simply as a message with a choice that the user can choose to do the command or not.

**Search terms**

status commands

ckipp01:
I'm going to probably play around with this for a while to see if I like
it before merging, but the idea here is that in situations where
something goes wrong it might be hard to tell because you only get the
small status message. Now you'll actually get a select that will show
you the tooltip and the command if there is one, or it will just warn
you of what the tooltip says.

Fixes #102

### DIFF
--- a/lua/metals.lua
+++ b/lua/metals.lua
@@ -183,29 +183,7 @@ M.info = function()
   end
 end
 
-M.toggle_logs = function()
-  local bufs = api.nvim_list_bufs()
-
-  for _, buf in ipairs(bufs) do
-    local buftype = api.nvim_buf_get_option(buf, "buftype")
-    local _, purpose = pcall(api.nvim_buf_get_var, buf, "metals_buf_purpose")
-
-    if buftype == "terminal" and purpose == "logs" then
-      local first_window_id = fn.win_findbuf(buf)[1]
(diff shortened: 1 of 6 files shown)


ckipp01:
I got no feedback against this, so I'll probably just go ahead and merge this to see if anyone complains. A bit of trial by fire I suppose 🔥 .

PR #366, opened 2022-04-16
Fix outdated link in MetalInfo

lenguyenthanh:
The gitter link seems return 404. so I removed it and also added discord link in MetalInfo table.

### DIFF
--- a/lua/metals.lua
+++ b/lua/metals.lua
@@ -156,7 +156,7 @@ M.info = function()
     end
     table.insert(output, "")
     table.insert(output, "## Helpful links")
-    table.insert(output, "  - https://gitter.im/scalameta/metals-vim")
+    table.insert(output, "  - https://discord.gg/FaVDrJegEh")
     table.insert(output, "  - https://github.com/scalameta/nvim-metals")
     table.insert(output, "  - https://github.com/scalameta/metals")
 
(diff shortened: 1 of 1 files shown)


lua/metals.lua
```diff
@@ -156,7 +156,7 @@ M.info = function()
     end
     table.insert(output, "")
     table.insert(output, "## Helpful links")
-    table.insert(output, "  - https://gitter.im/scalameta/metals-vim")
+    table.insert(output, "  - https://discord.gg/FaVDrJegEh")
```
ckipp01, line 159: Do you know if this link is one that expires after a certain amount of time, or is this one that was created under "expires never"? We want to make sure we're not including one that will expire.

Also, can we add another entry with the matrix bridge? So just add another line pointing to the matrix link.

https://matrix.to/#/#scalameta/vim-users:matrix.org

ckipp01 (changes requested):
Thanks for sending this in! Just a quick note on the matrix link.

lua/metals.lua
lenguyenthanh: Added matrix: https://matrix.to/#/#scalameta:vim-users

ckipp01 (approved):
Thanks a lot for the update @lenguyenthanh!

LGTM!

PR #400, opened 2022-05-25
feat: allows toggle logs to be horizontal or vertical

### LINKED ISSUE #250: Configurable position of the window with metals logs
**Describe the feature**

The command `:MetalsToggleLogs` creates a new vertical window. But, sometimes it's better to open logs in a horizontal window.  A new option to specify where exactly the window with logs should be opened could be useful.

Add a new option like `g:metals_logs_window = 'horizontal' | 'vertical'`. Or, maybe, it could have more options like `'top' | 'bottom' | 'left' | 'right'`.

**Additional context**

 This is hardcoded (according to discord) here:

> https://github.com/scalameta/nvim-metals/blob/ab78a00c1db8c8b189d0dcbe43768529edede15d/lua/metals.lua#L178 and https://github.com/scalameta/nvim-metals/blob/ab78a00c1db8c8b189d0dcbe43768529edede15d/lua/metals.lua#L186

**Search terms**

settings logs window

ckipp01:
This makes a couple small changes that allows the toggle logs command to
be able to take a param to either open vertically or horizontally. It
now takes a direction of either `vsp` or `sp` with a default of `vsp`.
This also makes a small change that when we create the user commands we
pass in whether or not an arg is expected, and you can set the amount of
args in the commands table.

Closes #250

### DIFF
--- a/doc/metals.txt
+++ b/doc/metals.txt
@@ -910,9 +910,12 @@ super_method_hierarchy()  Use to execute a |metals.super-method-hierarchy|
 switch_bsp()              Use to execute a |metals.bsp-switch| command.
 
                                                                  *toggle_logs()*
-toggle_logs()             Use to trigger a |tail -f .metals/log| on your
+toggle_logs({direction}) Use to trigger a |tail -f .metals/log| on your
                           current workspace that will open in the embedded
                           Neovim terminal.
+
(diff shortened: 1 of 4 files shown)


ckipp01:
No idea what is going on with CI.... this shouldn't affect those tests and locally I'm having no issue.

PR #419, opened 2022-06-26
Fix typos in code comments

kluen:
Just noticed these while reading the tvp code :)


ckipp01 (approved):
Thanks @kluen! ha, have fun reading all that. I'd love to actually re-write it all. It was right when I was learning about metatables, so I'm really happy it works, but it's probably not the prettiest 😆

kluen:
I'm completely new to lua (and metals concepts), so I wouldn't notice :D Thanks for your great work btw.

PR #422, opened 2022-06-28
add opts to hover_worksheet

antosha417:
I want to be able to override default border in hover worksheet floating window like this
`:lua require"metals".hover_worksheet {border = 'rounded'}`

I also did a little wrapper so if `hover_worksheet` is called twice than it will focus floating window (like regular `hover` behaviour). As I often find myself wanting either to scroll this floating window or to copy something from it. But I'm not sure whether you would want this in nvim-metals codebase or not.
```lua
local function hover_worksheet()
  local existing_float = vim.F.npcall(vim.api.nvim_buf_get_var, 0, 'lsp_floating_preview')
  if existing_float and vim.api.nvim_win_is_valid(existing_float) then
    vim.cmd('call win_gotoid(b:lsp_floating_preview)')
    return
  end
  metals.hover_worksheet { border = 'rounded' }
end

map('n', '<leader>hw', hover_worksheet)
```

### DIFF
--- a/doc/metals.txt
+++ b/doc/metals.txt
@@ -182,7 +182,7 @@ _ones that are sent and used by the server_
   * |showInferredType|
   * |superMethodLensesEnabled|
 
-_ones that are used internally by `nvim-metals`_ 
+_ones that are used internally by `nvim-metals`_
   * |disabledMode|
   * |decorationColor|
   * |serverOrg|
@@ -439,8 +439,8 @@ environment variable `JAVA_OPTS` and the `.jvmopts` file are respected.
 However, we care most about the proxy settings from those places, not
 necessarily the memory options since Metals has different requirements for
 example than sbt. So if you have proxy settings defined in `JAVA_OPTS`, then
-no need to define them here again. However, if you want to set some specific 
-memory settings for Metals, then this is the place. 
+no need to define them here again. However, if you want to set some specific
+memory settings for Metals, then this is the place.
 
 superMethodLensesEnabled                              *superMethodLensesEnabled*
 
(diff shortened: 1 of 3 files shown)


ckipp01 (changes requested):
Thanks for the pr @antosha417! So this is actually exposed here:

https://github.com/scalameta/nvim-metals/blob/main/lua/metals.lua#L434

So users never have to `require("metals.decoration")` but rather just `require("metals")` so we'll want to actually have `opts` available there as well and then pass it to this.

Also when do you that, can you update the `metals.txt` to show that `hover_worksheet` now takes an optional param?

ckipp01:
> I also did a little wrapper so if hover_worksheet is called twice than it will focus floating window (like regular hover behaviour). As I often find myself wanting either to scroll this floating window or to copy something from it. But I'm not sure whether you would want this in nvim-metals codebase or not.

Personally I don't think we need to add this since there is a fairly easy way to do this without adding this. For example after you issue the worksheet hover you can use a window command to just jump into it (CTRL-W  w) for example.

ckipp01 (approved):
Awesome, thanks for adding this @antosha417!

LGTM!

antosha417:
>  you can use a window command to just jump into it (CTRL-W w)

If I have `nvim-tree` window open then (CTRL-W w) or (CTRL-W CTRL-W) jumps there instead of hover window. So I need to do (CTRL-W W) with capital W, which is not so comfortable to press.
And I wanted it to feel very similar to `vim.lsp.buf.hover()` so that 
> Calling the function twice will jump into the floating window.

But it's completely ok for me to leave it in my own configuration. Or maybe we can add into some kind of wiki.

ckipp01:
Ah, good point. I don't use any sort of tree so I never hit on this. Feel free to stick it in discussions under "show and tell" https://github.com/scalameta/nvim-metals/discussions/new. Let me think on it a bit and yea maybe it is a good idea to add.

PR #437, opened 2022-07-18
fix empty strings as args passed always as arguments for neovim 0.7+

rochala:
Since neovim 0.7 this repository changed command creation into new api `api.nvim_create_user_command`.
It thou changed it's behavior and made `MetalsNew(Scala/Java)File` unusable as it now passed empty strings instead of nil when no arguments were given. It lead to funny behaviors for optional arguments which were present in 
`local new_file = function(language, directory_uri_opt, name_opt, file_type_opt)` implementation so directory_uri_opt on metals side was `Some("")` instead of None. It threw runtime error which now is fixed and all commands arguments are now nil when they equal `""`

This change in neovim API looks like is intended according to tests:
https://github.com/gpanders/neovim/blob/0d188f718b64aacd69f61eed54f6c47a1cb08414/test/functional/api/command_spec.lua#L192

### DIFF
--- a/lua/metals/config.lua
+++ b/lua/metals/config.lua
@@ -164,7 +164,7 @@ local function toggle_logs(direction)
   local bufs = api.nvim_list_bufs()
   local split_direction = "vsp"
 
-  if direction and direction ~= "" then
+  if direction then
     if direction ~= "vsp" and direction ~= "sp" then
       log.error_and_show("direction must be able vsp or sp")
       return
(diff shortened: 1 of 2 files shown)


ckipp01 (approved):
Hey @rochala thanks a lot for the pr! I hadn't noticed this broke since I'm using Telescope and it still works there. Just tested with both and this seems fine. Thanks for the fix! LGTM

PR #440, opened 2022-08-03
feat: add option to execute tvp node command on leaf toggle

antosha417:
This was proposed in [this](https://github.com/scalameta/nvim-metals/discussions/279#discussioncomment-3314378) discussion. And for me this option would make tvp feature more pleasant to use.

### DIFF
--- a/doc/metals.txt
+++ b/doc/metals.txt
@@ -988,6 +988,7 @@ The following configuration values are available to be passed into
     panel_alignment = "left",
     toggle_node_mapping = "<CR>",
     node_command_mapping = "r",
+    execute_node_command_on_leaf_toggle = false,
     collapsed_sign = "▸",
     expanded_sign = "▾",
   }
--- a/doc/metals.txt
+++ b/doc/metals.txt
@@ -988,7 +988,6 @@ The following configuration values are available to be passed into
     panel_alignment = "left",
     toggle_node_mapping = "<CR>",
     node_command_mapping = "r",
-    execute_node_command_on_leaf_toggle = false,
     collapsed_sign = "▸",
     expanded_sign = "▾",
   }
--- a/lua/metals/tvp/init.lua
+++ b/lua/metals/tvp/init.lua
@@ -14,7 +14,6 @@ local default_config = {
   panel_alignment = "left",
   toggle_node_mapping = "<CR>",
   node_command_mapping = "r",
-  execute_node_command_on_leaf_toggle = false,
   collapsed_sign = "▸",
   expanded_sign = "▾",
 }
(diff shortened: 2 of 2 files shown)


antosha417:
I'm not sure why the tests are failing :sweat:

ckipp01 (changes requested):
Thanks for the pr @antosha417, although I'm still not sure I want to do this. You mentioned in the discussion that this is common in other tree-like plugins. Could you give me some examples? Also if we do this, I don't think I'd want it as another setting. I try to avoid having configuration options and am really careful about adding new ones.

ckipp01:
> I'm not sure why the tests are failing 😓

Don't worry about the tests. There has been something funky going on lately that I'm unable to debug or haven't spent a lot of time on. As long as they are passing locally you're fine.

antosha417:
On the top of my head both [nerdtree](https://github.com/preservim/nerdtree) and [nvim-tree](https://github.com/kyazdani42/nvim-tree.lua) open folders or open files `o` or `<cr>`.

antosha417:
I've also noticed this behavior: When I close some of the split buffers and my cursor automatically goes to tvp buffer than this jump to last window doesn't work and the file opens in tvp panel.
```lua
    -- Jump to the last window so this doesn't open up in the actual tvp panel
    vim.cmd([[wincmd p]])
```
The easiest way to reproduce it is to just close all the windows except of tvp panel. But it can happen on other circumstances. 

So I'm thinking maybe in this case we want to create new window for it. With something like this:
```lua 
    -- Jump to the last window so this doesn't open up in the actual tvp panel
    vim.cmd([[wincmd p]])
    if vim.api.nvim_get_current_buf() == self.bufnr then
      local alignment = "topleft"
      if config.panel_alignment == "left" then
        alignment = "botright"
      end
      vim.cmd(string.format("silent %s vertical new ", alignment))
    end
```

What  do you think?

antosha417:
Also a lot of classes in my projects have companion objects. And they look the same it tvp buffer because they have the same name.
I see that metals has support for icons in tvp. Do you maybe want me to add some icons? 
What do you think about adding some highlight groups to the output so it will be a bit easier to read and distinguish different elements? 

Here for example we can easily see which `AbstractActor` is on object and which is a class:
Screen Shot 2022-08-04 at 0 14 41

ckipp01 (approved):
Alright, got around to playing around with this. I think it's non-invasive enough that this is totally fine. Thanks a lot for the PR! LGTM

ckipp01:
> I see that metals has support for icons in tvp. Do you maybe want me to add some icons?
> What do you think about adding highlight groups to the output so it will be a bit easier to read and distinguish different elements?

I would love having some icons.

PR #441, opened 2022-08-07
feat(tvp): add icons

antosha417:
This is how tvp with default icons looks for me:
Screen Shot 2022-08-07 at 18 55 45

### DIFF
--- a/doc/metals.txt
+++ b/doc/metals.txt
@@ -1008,8 +1008,8 @@ The following configuration values are available to be passed into
   }
 <
 
-In order for default icons to be displayed properly you need to install
-patched font.
+In order for the default icons to be displayed properly you need to have
+installed a patched font.
 
 The two main functions that you'll want to make sure to have mapped are
 |toggle_tree_view| and |reveal_in_tree|. When creating a mapping for any of
--- a/doc/metals.txt
+++ b/doc/metals.txt
@@ -990,8 +990,26 @@ The following configuration values are available to be passed into
     node_command_mapping = "r",
     collapsed_sign = "▸",
     expanded_sign = "▾",
+    icons = {
+      enabled = false,
+      symbols = {
+        object = "",
+        trait = "",
+        class = "ﭰ",
+        interface = "",
+        val = "",
+        var = "",
+        method = "ﬦ",
+        enum = "",
+        field = "",
+        package = "",
+      },
+    },
   }
 <
+
+In order for default icons to be displayed properly you need to install
+patched font.
 
 The two main functions that you'll want to make sure to have mapped are
 |toggle_tree_view| and |reveal_in_tree|. When creating a mapping for any of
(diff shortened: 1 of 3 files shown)


doc/metals.txt
```diff
@@ -990,9 +990,33 @@ The following configuration values are available to be passed into
...
+        val = "",
+        field = "",
+        method = "ﬦ",
+        constructor = "",
+        object = "",
+        class = "ﭰ",
+        trait = "",
+        interface = "",
+        package = "",
+        package_object = "",
```
antosha417, line 1005: I see that the name of an icon is of type `Kind` which is defined here https://github.com/scalameta/scalameta/blob/main/semanticdb/semanticdb/semanticdb.proto#L246-L264

But I didn't see any of `package_object`, `macro`, `type`, `parameter`, `self_parameter`, `type_parameter` or `local` icons in the tvp. So maybe it is a good idea to delete them from here.

doc/metals.txt
```diff
@@ -990,9 +990,33 @@ The following configuration values are available to be passed into
     node_command_mapping = "r",
     collapsed_sign = "▸",
     expanded_sign = "▾",
+    icons = {
+      enabled = false,
```
antosha417, line 994: It is disabled by default because I don't know how common is to have patched font installed.

doc/metals.txt
ckipp01: Yea, actually you don't need quite a few of these. There is a minimal set that are used. You can actually see them all [here](https://github.com/scalameta/metals/blob/main/metals/src/main/scala/scala/meta/internal/tvp/ClasspathTreeView.scala#L98-L111). So you should be able to simplify this to those.

doc/metals.txt
ckipp01: Yea, probably a good idea. After this is merged I can update the features list with a new screen shot showing this and with a message as a reminder that if someone is using a patched font they can enable this.

lua/metals/tvp/init.lua
```diff
@@ -16,6 +16,27 @@ local default_config = {
   node_command_mapping = "r",
   collapsed_sign = "▸",
   expanded_sign = "▾",
+  icons = {
```
ckipp01, line 19: See my comment in the docs, but this can be simplified to what I linked there.

ckipp01 (changes requested):
This looks awesome @antosha417! Thanks a lot for the pr. I linked a place where you can see the full list of icons. Let's match that, and then this should be good to merge.

ckipp01:
Also, tests are fixed now. So you should be able to rebase to see them 🟢 .

doc/metals.txt
antosha417: awesome, thanks!

ckipp01 (approved):
Looks great @antosha417! Thanks a lot for the PR! I added a small commit on top just to fix some typos in the docs.

LGTM!

PR #452, opened 2022-08-27
feat: allow user to override config.cmd

ckipp01:
NOTE: That when doing this, you're 100% on your own with how Metals gets
started. If what you pass in doesn't work, then nothing will work. Also
keep in mind that no java_opts or `config.server_opts` that you pass in
will be applied. You'll need to apply them yourself in the `cmd` table.

This isn't added to the docs because if you're sane you shouldn't do
this. This is for Anton.


keynmol (approved):
And Anton doth approve of this :D

PR #455, opened 2022-09-05
feat: add in metalsBinaryPath

ckipp01:
Another feature that should be useful for Nix users. In the situation where
they don't want nvim-metals to control the install, but also don't want to
use a global executable, this will allow them to instead use `metalsBinaryPath`
which allows you to pass in any path.

Know that you can't use this _with_ `useGlobalExecutable`. If you do, it will take
priority and win.

Addresses #454

### DIFF
--- a/doc/metals.txt
+++ b/doc/metals.txt
@@ -186,6 +186,7 @@ _ones that are sent and used by the server_
 _ones that are used internally by `nvim-metals`_
   * |disabledMode|
   * |decorationColor|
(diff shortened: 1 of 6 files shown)


gvolpe:
Thank you for doing this so quickly @ckipp01 , hope I can get this working soon! 🥳

ckipp01:
> Thank you for doing this so quickly @ckipp01 , hope I can get this working soon! 🥳

Let me know if you get stuck on anything!

PR #471, opened 2022-10-17
fix: scape % before setting metals_status

### LINKED ISSUE #468: `g:metals_status` breaks statusline because of unescaped characters
### Describe the bug

`g:metals_status` sometimes contains unescaped `%` symbols that breaks the status line evaluation (vim.api.nvim_eval_statusline) because those are special characters.

### Expected behavior

`%` and any other posible special characters according to [the statusline definition](https://neovim.io/doc/user/options.html#'statusline') should be escaped

### Operating system

macOS

### Version of Metals

v0.11.8

### Commit of nvim-metals

1284bbf8d79fe010909e65abdd849f047ff51914

caenrique:
Fixes #468

### DIFF
--- a/lua/metals/status.lua
+++ b/lua/metals/status.lua
@@ -2,7 +2,9 @@ local log = require("metals.log")
 
 -- @param status (string) a new status to be displayed
 local function set_status(status)
-  vim.api.nvim_set_var("metals_status", status)
+  -- Scaping the status to prevent % characters breaking the statusline
+  local scaped_status = status:gsub('[%%]', '%%%1')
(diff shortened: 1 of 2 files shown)


ckipp01 (changes requested):
Thanks for this fix @caenrique! I've been meaning to get around to it, but life's been getting in the way. Would you actually be able to write a small test for this in `status_spec.lua`. You should be able to just add another section to the existing test and make sure there is a `%` in it and that it shows what you'd expect at the end. Let me know if you have any questions about this.

caenrique:
done! :)

ckipp01 (approved):
Great, thanks @caenrique!

PR #472, opened 2022-10-17
feat: migrate to vim.lsp.start

### LINKED ISSUE #413: Migrate to `vim.lsp.start()`
### Task

In https://github.com/neovim/neovim/pull/18631 `vim.lsp.start()` was added which is basically an easier way to manage the attaching of a server than using `vim.lsp.start_client` since you don't need to keep track of whether or not a server has been attached to the current buffer with the same name and same root. This could get rid of some of the custom log we have for that here:

https://github.com/scalameta/nvim-metals/blob/6738fe3c0e2142de01c753f8b7c18281d11488ce/lua/metals/setup.lua#L87-L98

caenrique:
Hey! I was looking at some parts of the code and opened #471, and then I was curious about the `vim.lsp.start` call and started playing with it, and since it was pretty straight forward to have it working I've also opened this PR.

Hope it's close to how you was thinking of implementing it :)

Closes #413

### DIFF
--- a/lua/metals.lua
+++ b/lua/metals.lua
@@ -323,7 +323,6 @@ M.restart_server = function()
   end
 
   vim.defer_fn(function()
-    setup.reset_lsps()
     setup.initialize_or_attach()
   end, 3000)
 end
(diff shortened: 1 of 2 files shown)


ckipp01 (approved):
Thanks @caenrique! This is perfect. Exactly what I was thinking. It's super nice to simplify some of this stuff.

PR #473, opened 2022-10-18
feat: set virtual text highlight mode to combine

suliatis:
The effect of this change that the cursorline applies to the virtual
  text too in worksheet mode.

Before:
Screenshot 2022-10-18 at 8 18 06

After:
Screenshot 2022-10-18 at 8 05 38

I'm not sure if it ok to set this by default, it looks better imho, but I'm open to make this configurable.


ckipp01 (approved):
I think this is totally fine, looks good! No need for a configuration setting unless we get complaints. I try to avoid adding more configuration. LGTM, thanks!

btw, great minimal looking setup you have 💯

suliatis:
thank you

PR #474, opened 2022-10-20
feat: add test explorer

antosha417:
Hey! 
I've been wanting to run individual tests in `nvim-metals` since forever. Finally found some time and tried to do it with test explorer.
Seems like everything is working pretty good for me. 

I added two commands:
-  One for browsing all test suites and running one of them.
- Another one for browsing test suites only in current file and running one test case from selected suite.

Bouth commands can be rerun with `:lua require'dap'.run_last()`.
 
 In order to use it one need to edit metals config like this:
  ```lua
 local metals_config = metals.bare_config()
metals_config.settings = {
  ...
  testUserInterface = "Test Explorer",
}
metals_config.init_options.testExplorerProvider = true
```

Here is a small video of how it works:

https://user-images.githubusercontent.com/14187674/197073732-6696682b-1686-430f-87c0-1dac0844060f.mp4

### DIFF
--- a/lua/metals.lua
+++ b/lua/metals.lua
@@ -9,6 +9,7 @@ local install = require("metals.install")
 local log = require("metals.log")
 local messages = require("metals.messages")
 local setup = require("metals.setup")
+local test_explorer = require("metals.test_explorer")
 local util = require("metals.util")
 
 local has_plenary, Float = pcall(require, "plenary.window.float")
@@ -463,6 +464,14 @@ M.toggle_setting = function(setting)
 
     lsp.buf_notify(0, "workspace/didChangeConfiguration", { settings = { metals = settings } })
   end
+end
+
+M.select_test_suite = function()
+  test_explorer.dap_select_test_suite(conf.get_config_cache().root_dir)
+end
+
+M.select_test_case = function()
+  test_explorer.dap_select_test_case(conf.get_config_cache().root_dir)
 end
 
 return M
--- a/lua/metals.lua
+++ b/lua/metals.lua
@@ -467,11 +467,11 @@ M.toggle_setting = function(setting)
 end
 
 M.select_test_suite = function()
-  test_explorer.dap_select_test_suite(conf.get_config_cache().root_dir)
+  test_explorer.dap_select_test_suite()
 end
 
 M.select_test_case = function()
-  test_explorer.dap_select_test_case(conf.get_config_cache().root_dir)
+  test_explorer.dap_select_test_case()
 end
 
 return M
--- a/lua/metals/commands.lua
+++ b/lua/metals/commands.lua
@@ -213,6 +213,16 @@ local commands_table = {
     label = "Update Metals",
     hint = "Update to the latest Metals.",
   },
+  {
+    id = "select_test_suite",
+    label = "Select Test Suite",
+    hint = "Select test suite to run.",
+  },
+  {
+    id = "select_test_case",
+    label = "Select Test Case",
+    hint = "Select test case from current file to run.",
+  },
 }
 
 return {
--- a/lua/metals/commands.lua
+++ b/lua/metals/commands.lua
@@ -216,7 +216,7 @@ local commands_table = {
   {
     id = "select_test_suite",
     label = "Select Test Suite",
-    hint = "Select test suite to run.",
+    hint = "Select the test suite in your workspace to run.",
   },
   {
     id = "select_test_case",
--- a/lua/metals/commands.lua
+++ b/lua/metals/commands.lua
@@ -213,6 +213,9 @@ local commands_table = {
     label = "Update Metals",
     hint = "Update to the latest Metals.",
   },
+}
+
+local dap_commands_table = {
   {
     id = "select_test_suite",
     label = "Select Test Suite",
@@ -224,6 +227,11 @@ local commands_table = {
     hint = "Select test case from current file to run.",
   },
 }
+
+local dap_available, _ = pcall(require, "dap")
+if dap_available then
+  vim.list_extend(commands_table, dap_commands_table)
+end
 
 return {
   commands_table = commands_table,
--- a/lua/metals/config.lua
+++ b/lua/metals/config.lua
@@ -93,6 +93,7 @@ end
 local metals_init_options = {
   compilerOptions = {},
   debuggingProvider = debugging_provider,
+  testExplorerProvider = debugging_provider,
   decorationProvider = true,
   didFocusProvider = true,
   disableColorOutput = true,
--- a/lua/metals/config.lua
+++ b/lua/metals/config.lua
@@ -130,6 +130,7 @@ local valid_metals_settings = {
   "showImplicitConversionsAndClasses",
   "showInferredType",
   "superMethodLensesEnabled",
+  "testUserInterface",
 }
 
 -- We keep these seperated from the `valid_metals_settings` just for clarity.
--- a/lua/metals/messages.lua
+++ b/lua/metals/messages.lua
@@ -45,7 +45,17 @@ update your Metals executable. If you'd like nvim-metals to handle this for
 you, remove the `useGlobalExecutable` setting and try again.]]
 
 M.setup_dap_without_nvim_dap = [[
-You can call require("metals").setup_dap() without `nvim-dap` being installed.
+You can not call require("metals").setup_dap() without `nvim-dap` being installed.
+Please make sure `mfussenegger/nvim-dap` is installed.]]
+
+M.no_test_suites_found = [[
+No test sutes could be found in current file.]]
+
+M.no_test_cases_found = [[
+No test cases found in selected suite. Fallback to run whose suite.]]
+
+M.run_test_without_nvim_dap = [[
+You can not run test without `nvim-dap` being installed.
 Please make sure `mfussenegger/nvim-dap` is installed.]]
 
 return M
--- a/lua/metals/messages.lua
+++ b/lua/metals/messages.lua
@@ -52,7 +52,7 @@ M.no_test_suites_found = [[
 No test sutes could be found in current file.]]
 
 M.no_test_cases_found = [[
-No test cases found in selected suite. Fallback to run whose suite.]]
+No test cases found in selected suite. Fallback to run whole suite.]]
 
 M.run_test_without_nvim_dap = [[
 You can not run test without `nvim-dap` being installed.
--- a/lua/metals/messages.lua
+++ b/lua/metals/messages.lua
@@ -49,7 +49,7 @@ You can not call require("metals").setup_dap() without `nvim-dap` being installe
 Please make sure `mfussenegger/nvim-dap` is installed.]]
 
 M.no_test_suites_found = [[
-No test sutes could be found in current file.]]
+No test suites could be found in current file.]]
 
 M.no_test_cases_found = [[
 No test cases found in selected suite. Fallback to run whole suite.]]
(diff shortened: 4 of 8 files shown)

### CODE REVIEW

**Comment by ckipp01** [comment]
Very cool @antosha417! Thanks so much for digging into this! I'll do a more in depth review in a bit, but I've just been trying to use it first. A couple things I notice right away. Without setting the following in the `metals_init_options` it will never return any suites:

```lua
  testExplorerProvider = true,
```
Was this just an oversight? If not, I'm not sure how it was working in your gif since mine will never get a `metals-update-test-explorer` without that being set.

Once I do have that set and I start using it I always end up with:
```
LSP[metals] Target 'file:/Users/ckipp/Documents/scala-workspace/scalameta-org/metals/?id=unit-test' not found.
```
No matter what target I'm in. I've tried in a fresh Hello World project and I've also tried in the Metals code base as well. The above was trying to run a test in `tests/unit/src/test/scala/tests/ScalaVersionsSuite.scala` in the Metals code base. I'm using the latest snapshot: `0.11.9+56-6943e41c-SNAPSHOT` and in the logs I'm seeing:

```
SEVERE: Internal error: scala.meta.internal.metals.debug.BuildTargetNotFoundException: Build target not found: file:/Users/ckipp/Documents/scala-workspace/scalameta-org/metals/?id=unit-test
java.util.concurrent.CompletionException: scala.meta.internal.metals.debug.BuildTargetNotFoundException: Build target not found: file:/Users/ckipp/Documents/scala-workspace/scalameta-org/metals/?id=unit-test
        at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:332)
        at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:347)
        at java.base/java.util.concurrent.CompletableFuture$UniAccept.tryFire(CompletableFuture.java:708)
        at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
        at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2162)
        at scala.concurrent.java8.FuturesConvertersImpl$CF.apply(FutureConvertersImpl.scala:29)
        at scala.concurrent.java8.FuturesConvertersImpl$CF.apply(FutureConvertersImpl.scala:26)
        at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:484)
        at scala.concurrent.ExecutionContext$parasitic$.execute(ExecutionContext.scala:222)
        at scala.concurrent.impl.Promise$Transformation.submitWithValue(Promise.scala:429)
        at scala.concurrent.impl.Promise$DefaultPromise.submitWithValue(Promise.scala:338)
        at scala.concurrent.impl.Promise$DefaultPromise.tryComplete0(Promise.scala:285)
        at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:504)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
        at java.base/java.lang.Thread.run(Thread.java:833)
```
So for some reason Metals isn't able to find the build target. Not sure if it's something funky going on with the URI or what, but I'll play around. 

Really excited about this!

lua/metals/test_explorer.lua
```diff
@@ -0,0 +1,178 @@
...
+        },
+        jvmOptions = {},
+        environmentVariables = {},
+      },
+    },
+  }
+  dap.run(arguments)
+end
+
+M.dap_select_test_suite = function(root)
```
ckipp01, line 119: So for these I don't think we even need root. Because in `dap_run_test` we end up using that to put together the `uri` when actually we don't want to do that as it doesn't seem to work. Instead when the `executeClientCommand` comes in with the `metals-update-test-explorer` there is a `targetUri` that is sent with each target. Instead of using `root` here we need to use the `targetUri` instead.

lua/metals.lua
```diff
@@ -465,4 +466,12 @@ M.toggle_setting = function(setting)
   end
 end
 
+M.select_test_suite = function()
+  test_explorer.dap_select_test_suite(conf.get_config_cache().root_dir)
```
ckipp01, line 470: I mentioned this in the other file, but I don't think we actually need to include the `root_dir` here. Instead we should rely on the `targetUri` that comes with the `metals-update-test-explorer` command.

lua/metals.lua
```diff
@@ -465,4 +466,12 @@ M.toggle_setting = function(setting)
...
+end
+
+M.select_test_case = function()
+  test_explorer.dap_select_test_case(conf.get_config_cache().root_dir)
```
ckipp01, line 474: Same as above no need for the `root_dir` here.

lua/metals/commands.lua
```diff
@@ -213,6 +213,16 @@ local commands_table = {
     label = "Update Metals",
     hint = "Update to the latest Metals.",
   },
+  {
+    id = "select_test_suite",
+    label = "Select Test Suite",
+    hint = "Select test suite to run.",
```
ckipp01, line 219:
```suggestion
    hint = "Select the test suite in your workspace to run.",
```
What do you think about expanding it a bit to let the user know that this is choosing from the full workspace?

lua/metals/config.lua
```diff
@@ -130,6 +130,7 @@ local valid_metals_settings = {
   "showImplicitConversionsAndClasses",
   "showInferredType",
   "superMethodLensesEnabled",
+  "testUserInterface",
```
ckipp01, line 133: I'll comment here because I can't up above. But I think we also need to add `testExplorerProvider = debugging_provider` like we do with `debuggingProvider`. This way it will ensure dap is available and turn this on when it is.

lua/metals/commands.lua
```diff
@@ -213,6 +213,16 @@ local commands_table = {
     label = "Update Metals",
     hint = "Update to the latest Metals.",
   },
+  {
```
ckipp01, line 216: So one thing I was thinking about is right now these are unconditionally added. If a user doesn't have DAP setup they'll still see these. We probably don't want that right?

lua/metals/messages.lua
```diff
@@ -45,7 +45,17 @@ update your Metals executable. If you'd like nvim-metals to handle this for
 you, remove the `useGlobalExecutable` setting and try again.]]
 
 M.setup_dap_without_nvim_dap = [[
-You can call require("metals").setup_dap() without `nvim-dap` being installed.
+You can not call require("metals").setup_dap() without `nvim-dap` being installed.
+Please make sure `mfussenegger/nvim-dap` is installed.]]
+
+M.no_test_suites_found = [[
+No test sutes could be found in current file.]]
```
ckipp01, line 52:
```suggestion
No test suites could be found in current file.]]
```

lua/metals/messages.lua
```diff
@@ -45,7 +45,17 @@ update your Metals executable. If you'd like nvim-metals to handle this for
...
+
+M.no_test_cases_found = [[
+No test cases found in selected suite. Fallback to run whose suite.]]
```
ckipp01, line 55:
```suggestion
No test cases found in selected suite. Fallback to run whole suite.]]
```

antosha417:
Yeah, I completely forgot that I also did `metals_config.init_options.testExplorerProvider = true` for it to work. I added it at the top. 

Regarding the target uri I cheated a little and just took `project_root .. '/?id=target_name'. Because test explorer just sends the `target_name` not the uri.   
I think I can request this uri from metals. As we do in `show_build_target_info`. Will try to fix it soon.

lua/metals/test_explorer.lua
```diff
@@ -0,0 +1,178 @@
...
+  if event.kind == "removeSuite" then
+    handle_remove_suite(targetName, event)
+  elseif event.kind == "addSuite" then
+    handle_add_suite(targetName, event)
+  elseif event.kind == "updateSuiteLocation" then
+    handle_update_suite_location(targetName, event)
+  elseif event.kind == "addTestCases" then
+    handle_add_test_cases(targetName, event)
+  else
+    log.error("Unknown event in test explorer", targetName, event)
```
ckipp01, line 61: Do you think we should surface this to the user with `error_and_show`? wdyt?

ckipp01 (changes requested):

ckipp01:
> Regarding the target uri I cheated a little and just took `project_root .. '/?id=target_name'`. Because test explorer just sends the `target_name` not the uri. I think I can request this uri from metals. As we do in `show_build_target_info`. Will try to fix it soon.

Yea you shouldn't actually have to request anything extra at all. Here's an example:

```
[Trace - 02:42:41 pm] Sending notification 'metals/executeClientCommand'
Params: {
  "command": "metals-update-test-explorer",
  "arguments": [
    {
      "targetName": "minimal.test",
      "targetUri": "file:/Users/ckipp/Documents/scala-workspace/minimal/minimal/test/?id\u003dminimal.test",
      "events": [
        {
          "fullyQualifiedClassName": "example.HelloSuite",
          "className": "HelloSuite",
          "symbol": "example/HelloSuite#",
          "location": {
            "uri": "file:///Users/ckipp/Documents/scala-workspace/minimal/minimal/test/src/example/HelloSuite.scala",
            "range": {
              "start": {
                "line": 4,
                "character": 6
              },
              "end": {
                "line": 4,
                "character": 16
              }
            }
          },
          "canResolveChildren": true,
          "kind": "addSuite"
        },
        {
          "fullyQualifiedClassName": "example.HelloSuite",
          "className": "HelloSuite",
          "testCases": [
            {
              "name": "numbers",
              "location": {
                "uri": "file:///Users/ckipp/Documents/scala-workspace/minimal/minimal/test/src/example/HelloSuite.scala",
                "range": {
                  "start": {
                    "line": 5,
                    "character": 2
                  },
                  "end": {
                    "line": 5,
                    "character": 6
                  }
                }
              }
            },
            {
              "name": "reproduce",
              "location": {
                "uri": "file:///Users/ckipp/Documents/scala-workspace/minimal/minimal/test/src/example/HelloSuite.scala",
                "range": {
                  "start": {
                    "line": 14,
                    "character": 2
                  },
                  "end": {
                    "line": 14,
                    "character": 6
                  }
                }
              }
            }
          ],
          "kind": "addTestCases"
        }
      ]
    }
  ]
}
```
You can grab the `targetUri` right from there.

ckipp01 (commented):
Working like a charm @antosha417! One final thing I'd like to have before merging. Can we also add a small section into `:h metals-nvim-dap` about this? Something telling people that they actually have to change a setting to get this to work and that once set they will no longer see code lenses for running/debugging.

antosha417:
>  once set they will no longer see code lenses for running/debugging 

Actually I think it would be nice to show code lenses near the test names. So users will be able to run them and execute specific test cases or whose test suites. 
Currently in `test_explorer.state` we have info about the location of tests that can be run. 

I'll read more about code lenses and see if I can implement it. What do you think?

antosha417:
I was able to set code lenses and even execute them. But sadly neovim requests them again from metals and overrides mine with empty list. 
So I think for code lenses it should be a feature in the metals itself to set them for each test case.

ckipp01 (approved):
> Actually I think it would be nice to show code lenses near the test names.

Yea that's tricky though since the "Test Explorer" is sort of meant to be an alternative to code lenses. One other solution we could look into is instead of using code lenses for this is to use extmarks to mark the individual tests. They'd then be on the left hand side, but still useful to see. We could have full control over them and it was what I had in mind originally.

Thanks a lot for you work on this @antosha417!

LGTM!

PR #514, opened 2022-12-23
feat: add a restart command with langoustine

ckipp01:


ghostbuster91:
This would be awesome, I've literally just thought yesterday that such a tool would be handful, though I was rather thinking how to do it in the terminal.

PR #515, opened 2022-12-27
feat: pass mainClass in DAP request

### LINKED ISSUE #513: Allow specification of "mainClass" and "buildTarget" DAP keys
### Describe the feature

I work on a large project with many build targets. I am rarely working inside the one that contains the `main` I want to debug. I would like to be able to launch a configuration that runs a configuration with a specific `mainClass` no matter what Scala file I am editing. Currently, I need to switch to the file that contains `main` to start the debug session.

[I think it is possible to specify a main class](https://scalameta.org/metals/docs/integrations/debug-adapter-protocol/#via-explicit-main-or-test-commands) (VSCode supports this).

Without reading the source too carefully:

- Pass `mainClass` [here](https://github.com/scalameta/nvim-metals/blob/d1c01907256dae7c9d55ba1fcfb8cf6b4f583325/lua/metals/setup.lua#L108).
- I imagine something will need to change for configuration selection since these may not correspond to the current build target.

I'm going to take a shot at implementing this, so I'd appreciate any tips.

trmckay:
Closes #513. Allows specification of `mainClass` in debug configurations. Example:

```lua
require("dap").configurations.scala = {
    {
        type = "scala",
        name = "Debug with explicit main class",
        request = "launch",
        metals = {
            mainClass = "com.foo.App",
        },
    },
}
```

I've tested this locally and it works as I would expect it to.

### DIFF
--- a/lua/metals/setup.lua
+++ b/lua/metals/setup.lua
@@ -107,12 +107,16 @@ local function setup_dap(execute_command)
       local metals_dap_settings = config.metals or {}
       arguments = {
         path = vim.uri_from_bufnr(0),
-        runType = metals_dap_settings.runType or "run",
         args = metals_dap_settings.args,
         jvmOptions = metals_dap_settings.jvmOptions,
         env = metals_dap_settings.env,
         envFile = metals_dap_settings.envFile,
       }
+      if metals_dap_settings.mainClass then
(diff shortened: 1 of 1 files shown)


ckipp01 (approved):
Hey @trmckay, thanks for this! Yup, this is exactly what I had in mind.

LGTM!

PR #544, opened 2023-03-18
feat: Add `enableSemanticHighlighting` to valid metals settings

ghostbuster91:


lua/metals/config.lua
```diff
@@ -132,6 +132,7 @@ local valid_metals_settings = {
   "showInferredType",
   "superMethodLensesEnabled",
   "testUserInterface",
+  "enableSemanticHighlighting"
```
ckipp01, line 135: Total nit, I try to keep these in alphabetical order. Are you able to move this up?

ckipp01 (approved):
Thanks for the pr @ghostbuster91! Just a quick nit. You'll also need to run `make format` to make sure formatting is correct.

ckipp01 (changes requested):
Sorry hit the wrong button!

mpasa:
Is there a list of these groups specific to metals we can link in the documentation? I know of the ones like `@lsp.type.parameter` but I know Rust has `@lsp.type.struct.rust` for example.

lua/metals/config.lua
ghostbuster91: fixed

ckipp01:
> Is there a list of these groups specific to metals we can link in the documentation? I know of the ones like @lsp.type.parameter but I know Rust has @lsp.type.struct.rust for example.

Good question, I'm not actually 100% sure. I'd recommend asking in the general #metals channel on discord. I haven't been really involved/following the semantichighlighting stuff so I'm not 100% sure.

mpasa:
Tested and it works well for me. With `:Inspect`, I'm able to see things like:

```
Semantic Tokens
  - @lsp.type.class.scala links to Type 
  - @lsp.type.namespace.scala links to Type 
  ...
```

PR #548, opened 2023-04-24
refactor: make error for null in reveal in tree better

### LINKED ISSUE #546: Better message for 'null'-response from `metals/treeViewReveal`
### Describe the feature

When `metals/treeViewReveal` responds with `null`, nvim-metals [logs](https://github.com/scalameta/nvim-metals/blob/0b9c5303b7402ae24971ed2f0e5890fd72505532/lua/metals/tvp/init.lua#L473) the following message: "You recieved a node for a view nvim-metals doesn't support"

I'd expect a message which either encourages to open an issue in Metal (is it a case, when we should report to Metals?) or clarifies the impossibility of this action due to Metals lacking some information

Reproduction steps:
- in a scala 3 project open `scala.quoted.Quotes`
- invoke `lua require('metals.tvp').reveal_in_tree()<CR>`
- look at lsp.trace.json
```
[Trace - 00:23:28 AM] Received request 'metals/treeViewReveal - (9)'
Params: {
  "textDocument": {
    "uri": "file:///..../.metals/readonly/dependencies/scala3-library_3-3.2.2-sources.jar/scala/quoted/Quotes.scala"
  },
  "position": {
    "line": 24,
    "character": 35
  }
}


[Trace - 00:23:28 AM] Sending response 'metals/treeViewReveal - (9)'. Processing request took 90ms
Result: null
```

_No response_

ckipp01:
This will hopefully give the user a bit more information about what is
happening when they receive a `null` from the tree view. Since it's not
supported yet in Scala 3 this may happen. So in that situation we tell
them that and also let them know that if they are in Scala 2 they should
create an issue. We also better catch the situation where they truly get
a view id that isn't supported. In that case we show them what id it is.

Closes #546

### DIFF
--- a/lua/metals/messages.lua
+++ b/lua/metals/messages.lua
@@ -58,4 +58,9 @@ M.run_test_without_nvim_dap = [[
 You can not run test without `nvim-dap` being installed.
 Please make sure `mfussenegger/nvim-dap` is installed.]]
 
+M.scala_3_tree_view = [[
+You recieved a null response for the tree view.
+If you're using Scala 3, this is expected as it's not support yet.
+If you're using Scala 3, please create an issue for this in nvim-metals.]]
+
 return M
--- a/lua/metals/tvp/init.lua
+++ b/lua/metals/tvp/init.lua
@@ -2,6 +2,7 @@ local api = vim.api
 local lsp = vim.lsp
 
 local log = require("metals.log")
+local messages = require("metals.messages")
 local Node = require("metals.tvp.node")
 local tvp_util = require("metals.tvp.util")
 local util = require("metals.util")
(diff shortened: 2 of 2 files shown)


lua/metals/messages.lua
```diff
@@ -58,4 +58,9 @@ M.run_test_without_nvim_dap = [[
 You can not run test without `nvim-dap` being installed.
 Please make sure `mfussenegger/nvim-dap` is installed.]]
 
+M.scala_3_tree_view = [[
+You recieved a null response for the tree view.
+If you're using Scala 3, this is expected as it's not support yet.
+If you're using Scala 3, please create an issue for this in nvim-metals.]]
```
susliko, line 64: Probably, Scala 2 is meant in the second sentence

ckipp01: whoops, thank you for catching this.

Issue #551, opened 2023-04-25
Better "Restart X" command names


susliko:
### Describe the feature

Caught myself always pausing on this step whenever I want to restart either Metals or build server. "Restart Server" means the restart of Metals, but build server is also a server :)

<img width="719" alt="image" src="https://user-images.githubusercontent.com/15569000/234231297-e4dacff2-8620-4faf-8abf-e15614a2379b.png">

### Potential ways to implement

Probably, command names like
"Restart Build Server" and "Restart LSP Server", or
"Restart Build" and "Restart Metals" might help to disambiguate

ckipp01:
ah yea these could be better. They were originally mimicking what the actual commands in Metals were, which aren't really clear. Feel free to PR a change to make these clearer.

ckipp01:
Closed via #552

PR #552, opened 2023-04-25
Better "Restart X" command names

susliko:
Addresses https://github.com/scalameta/nvim-metals/issues/551
`restart_server` is renamed to `restart_metals`
`restart_build` is renamed to `restart_build_server`


ckipp01 (changes requested):
Thanks @susliko! Just a nit, can you add a `BREAKING CHANGE: description` to the footer? Lots of nvim users will use packer and it picks up the various ways to indicate a breaking change using conventional commits and warns users when they update. Since this is technically a breaking change to the LUA API we'll want to make sure we do that here.

ckipp01 (approved):
Thanks @susliko !

PR #585, opened 2023-07-07
Tweaked find_root_dir to accept a configurable max # of grandparents

JL102:
With this change, users can now set `config.find_root_dir_max_project_nesting` to an integer if you want to override the default # of parents.

Related to discussion #584

Note: It doesn't have to use recursion, since it was already looping through a finite list of parent folders, one at a time.

Second note: I didn't write any tests since I'm not sure how to set those up. I did test it myself in my own project. **(Running `make tests` failed on my system.)** What I'd probably do is set up a folder tree like this:

testdir/
  - build.sbt
  - a/
    - build.sbt
    - b/
      - build.sbt
      - c/
        - build.sbt
        - src/main/scala/Main.scala

and set `config.find_root_dir_max_project_nesting` to values between 0 and 4, with the following expectations:
0: root_dir = `testdir/a/b/c/build.sbt`
1: root_dir = `testdir/a/b/build.sbt`
2: root_dir = `testdir/a/build.sbt`
3: root_dir = `testdir/build.sbt`
4: root_dir = `testdir/build.sbt` (same as 3 because it's the max)

Additionally, setting it to something like 20 would be useful to guarantee that it doesn't break when it reaches the filesystem root.

Third note: I'm pretty sure this won't break on systems where people already overwrote their own find_root_dir, since Lua seems to not care when you add extra parameters to the function (find_root_dir now takes `find_root_dir_max_project_nesting` as a third parameter, but people who overwrote it will not have a third parameter, so for them, it should just ignore the third parameter)

### DIFF
--- a/lua/metals/rootdir.lua
+++ b/lua/metals/rootdir.lua
@@ -11,7 +11,8 @@ local has_pattern = function(patterns, target)
   end
 end
 
---- NOTE: This only searches 2 levels deep to find nested build files.
+--- NOTE: maxParentSearch lets you check up to a certain number of parent
+--- folders to find nested build files.
 --- Given a situation like the below one where you have a root build.sbt
 --- and one in your module a, you want to ensure the root is correctly set as
 --- the root one, not the a one. This checks the parent dir to ensure this.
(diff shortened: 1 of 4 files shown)


ckipp01:
Thanks for this @JL102! I'll look it over in the coming days. I have some issues with the current tests, so let me fix that and then get to this.

lua/metals/rootdir.lua
```diff
@@ -11,32 +11,42 @@ local has_pattern = function(patterns, target)
...
   local path = Path:new(startpath)
   -- TODO if we don't find it do we really want to search / probably not... add a check for this
-  for _, parent in ipairs(path:parents()) do
+  -- First parent index in which we found a target file
+  local firstFoundIdx = -1
+  local ret = nil
+
+  for i, parent in ipairs(path:parents()) do
+    -- Exit loop before checking anything if we've exceeded the search limits
+    if firstFoundIdx >= 0 and (i - firstFoundIdx > maxParentSearch) then
```
ckipp01, line 35: Shouldn't this be

```suggestion
    if firstFoundIdx >= 0 and (i - firstFoundIdx >= maxParentSearch) then
```

For example if you open up a project like:

```
.scala-build/
.bsp/
project.scala
```

Which is a valid scala-cli workspace, and you don't have `maxParentSearch` set so it defaults to 1, with this being `>` you end up searching 2 layers down, when you only want to search one.

ckipp01 (changes requested):
I'm sorry this has sat here for so long @JL102. I have been using this the last couple days and just noticed one thing.

lua/metals/rootdir.lua
ckipp01: Actually to further simplify this, wouldn't this also work?

```lua
local find_root_dir = function(patterns, startpath, maxParentSearch)
  local path = Path:new(startpath)
  -- TODO if we don't find it do we really want to search / probably not... add a check for this
  -- First parent index in which we found a target file
  local firstFoundIdx = nil
  local ret = nil

  for i, parent in ipairs(path:parents()) do
    -- Exit loop before checking anything if we've exceeded the search limits
    if firstFoundIdx and (i > maxParentSearch) then
      return ret
    end
    local pattern = has_pattern(patterns, parent)
    if pattern then
      -- Mark the first parent that was found, so we can exit the loop when we've exhausted our search limits
      if not firstFoundIdx then
        firstFoundIdx = i
      end
      -- (over)write the return value with the highest parent found
      ret = parent
    end
  end
  -- In case we went through the entire loop (e.g. if maxParentSearch is really high)
  return ret
end
```

ckipp01 (approved):
Alright, I ended up just committing on top of this @JL102 to help it move along. I also added a test showing that it works, and all the existing tests still pass. I feel pretty confident in this so I'll go ahead and merge. Again, thanks for pushing this forward!

lua/metals/rootdir.lua
JL102: Re the first comment: The way I interpreted the setting is that the default behavior (which is nvim-metals' behavior prior to this change) searches *1 parent folder up*, and if build.sbt is in the same folder as project.scala then that's *0 parent folders up* because it's the same folder.
Re the second comment and code suggestion: I don't have access currently to the server that I was developing on, so I can't test it at the moment. But I trust your Lua skills, so if it's got the same behavior I'm cool with it!

PR #605, opened 2023-09-07
Update doc about enableSemanticHighlighting

ex0ns:
Following the discussion in https://github.com/scalameta/nvim-metals/discussions/603 I have opened this PR as a reminder to myself to take of that.
I am sure that we could update the description as well but I'm wondering if this documentation is taken from elsewhere, I also need to double check if any other options/default changed in order to include the changes in that PR.


ckipp01 (approved):
Hey! Thanks for this! I think this is fine for now. LGTM

Issue #617, opened 2023-10-11
Unable to run MetalsInfo on nightly


vanhtuan0409:
### Describe the bug

Unable to run MetalsInfo on nightly build  due to deprecated `vim.lsp.util._trim`

https://github.com/scalameta/nvim-metals/blob/main/lua/metals.lua#L176

### Expected behavior

Able to run MetalsInfo

### Operating system

Linux

### Version of Metals

v1.0.1

### Commit of nvim-metals

HEAD

PR #618, opened 2023-10-11
fix: deprecated vim.lsp.util._trim

vanhtuan0409:
fix: https://github.com/scalameta/nvim-metals/issues/617


ckipp01 (approved):
Thanks for the fix @vanhtuan0409! The CI is failing, but that's not due to this change. There is something funky going on with the decorations modules with nightly that I need to figure out. LGTM

PR #622, opened 2023-10-28
Add bsp status support

rochala:
Enables BSP status messages to be sent from the LSP server and stores them similarly to metals status text, being inside global variable "bsp_status".

In order to add it to your status line, it can be accessed by:
```lua
vim.g["bsp_status"] or ""
```

For some reason, the bsp status is not sent from the server when type is `show-message`.

### DIFF
--- a/lua/metals/config.lua
+++ b/lua/metals/config.lua
@@ -103,6 +103,7 @@ local metals_init_options = {
   inputBoxProvider = true,
   quickPickProvider = true,
   statusBarProvider = "show-message",
+  bspStatusBarProvider = "on",
   treeViewProvider = true,
 }
 
--- a/lua/metals/status.lua
+++ b/lua/metals/status.lua
@@ -4,7 +4,7 @@ local log = require("metals.log")
 local function set_status(status, type)
   -- Scaping the status to prevent % characters breaking the statusline
   local scaped_status = status:gsub("[%%]", "%%%1")
-  local status_var = type .. "_status"
+  local status_var = (type or "metals") .. "_status"
   vim.api.nvim_set_var(status_var, scaped_status)
 end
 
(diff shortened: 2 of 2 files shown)


lua/metals/config.lua
```diff
@@ -103,6 +103,7 @@ local metals_init_options = {
   inputBoxProvider = true,
   quickPickProvider = true,
   statusBarProvider = "show-message",
+  bspStatusBarProvider = "on",
```
ckipp01, line 106: By having this on by default will users that have `show-message` on lose out on any information? If I understand what you said in the PR description none of the bsp status stuff is being sent when `show-message` is on right? Is that intended? Is the bsp status message all only additional information? Did you try explicity setting it to `show-message` to see if anything happened? Ideally in my head when I first saw this I wanted to just abstract it away and not even expose it. Simply if the user has `statusBarProvider` turned `on`, then they'd get both. If it was set to `show-message`, then it was set for both. I don't really see the point of having them be seperate.

ckipp01 (commented):
Thanks a lot of this @rochala! Ignore the test failure, it's something with nightly and decorations that I haven't looked into yet. Just left one comment/question.

lua/metals/config.lua
rochala: The BSP status provider option has to be set to "on" in order to send both status value to display on status line and also commands to restart a BSP when it is not working.

Let's recap this here:
`bspStatusBar = "on"` = bsp status is accessible from `vim.g["bsp_status"]` and reconnect prompts are also displayed,
`bspStatusBar = "show-message"` = there are only prompts, and they are sent by `window/showMessageRequest` not by `metals/status` command field.

If a user has show-message set in his configuration, he will have empty global variable bsp_status and that's it. I have double-checked the commands, and the prompts are present on both "show-message" and "on"

If you want to quickly test it for yourself, the easiest way to force the build tool to stop is `kill -STOP pid`

ckipp01 (approved):
Hey, so sorry for the delay again on this. Thanks for the explanation. I'll go ahead and just merge as is. 👍🏼

PR #670, opened 2024-04-25
Log and show error if user_on_attach fails

ghostbuster91:
I had an error in my `on_attach` function that was causing it to silently fail before any mappings where created. There was no info at all about the failure which made it harder to understand what is happening.

This PR logs any errors that could happen during the execution of user's `on_attach` callback.


ckipp01 (approved):
Makes sense, thanks @ghostbuster91 !

ckipp01:
> @ckipp01 - will there be one more release which fixes the said issue

All you need to do is update the plugin, it will pull in this change.

PR #689, opened 2024-09-14
feat: add inlay hints setting

gAbelli:
Add support for inlay hints settings introduced in Metals v1.3.0.

Reference:

- https://scalameta.org/metals/blog/2024/04/15/thalium#inlay-hints-support
- https://scalameta.org/metals/docs/editors/user-configuration

### DIFF
--- a/doc/metals.txt
+++ b/doc/metals.txt
@@ -184,6 +184,7 @@ _ones that are sent and used by the server_
   * |excludedPackages|
   * |fallbackScalaVersion|
   * |gradleScript|
+  * |inlayHints|
   * |javaFormat.eclipseConfigPath|
   * |javaFormat.eclipseProfile|
   * |javaHome|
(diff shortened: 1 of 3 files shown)


ckipp01 (changes requested):
Thanks for adding this @gAbelli! Since we're moving this into `inlayHints` like they're supposed to be we can also remove the following:

```
    showImplicitArguments = true,
    showImplicitConversionsAndClasses = true,
    showInferredType = true,
```

They no longer need to be their own valid keys. While from the looks of it in Metals they _are_ still supported lets only advertise a single (new) way of doing this.

gAbelli:
I agree, I dropped them from the documentation but I kept them among the valid settings for backwards compatibility

ckipp01 (approved):
Looks good, thanks!

PR #693, opened 2024-10-05
Remove emphasis on 'not' in docs

ksonj:
I am getting an error on a duplicate tag `"not"` when generating helptags, due to the word being wrapped in asterisks here. I think it's not intended to tag on "not" here, so I've simply removed the emphasis.


ckipp01 (approved):
Thanks!

Issue #664, opened 2024-04-13
not found: "metals/findTextInDependencyJars" request handler for client "GitHub Copilot".


alycklama:
### Describe the bug

When executing `:MetalsFindInDependencyJars` on the command-line with the GitHub Copilot plugin installed with the following input:

```
Filemask: .conf
Query: String
```

I then run into the following error:

```bash
Error executing Lua callback: ...l/Cellar/neovim/0.9.5/share/nvim/runtime/lua/vim/lsp.lua:1382: not found: "metals/findTextInDependencyJars" request handler for client "GitHub Copilot".
stack traceback:
        [C]: in function 'assert'
        ...l/Cellar/neovim/0.9.5/share/nvim/runtime/lua/vim/lsp.lua:1382: in function 'request'
        ...l/Cellar/neovim/0.9.5/share/nvim/runtime/lua/vim/lsp.lua:1969: in function 'fn'
        ...l/Cellar/neovim/0.9.5/share/nvim/runtime/lua/vim/lsp.lua:184: in function 'for_each_buffer_client'
        ...l/Cellar/neovim/0.9.5/share/nvim/runtime/lua/vim/lsp.lua:1968: in function 'buf_request'
        ...lt.ext/.local/share/nvim/lazy/nvim-metals/lua/metals.lua:281: in function 'send_request'
        ...lt.ext/.local/share/nvim/lazy/nvim-metals/lua/metals.lua:292: in function 'on_confirm'
        ...al/Cellar/neovim/0.9.5/share/nvim/runtime/lua/vim/ui.lua:103: in function 'input'
        ...lt.ext/.local/share/nvim/lazy/nvim-metals/lua/metals.lua:288: in function 'get_query_and_send'
        ...lt.ext/.local/share/nvim/lazy/nvim-metals/lua/metals.lua:302: in function 'on_confirm'
        ...al/Cellar/neovim/0.9.5/share/nvim/runtime/lua/vim/ui.lua:103: in function 'input'
        ...lt.ext/.local/share/nvim/lazy/nvim-metals/lua/metals.lua:297: in function <...lt.ext/.local/share/nvim/lazy/nvim-metals/lua/metals.lua:276>
        .../.local/share/nvim/lazy/nvim-metals/lua/metals/setup.lua:46: in function <.../.local/share/nvim/lazy/nvim-metals/lua/metals/setup.lua:45>
```

Why is GitHub Copilot affected by a command for Metals? Any idea what could be going on here?

[...]

PR #697, opened 2024-12-06
fix: ignore responses from other clients

ckipp01:
There was a change a while back that makes the buf request go to all the
attached clients. For Scala normally this isn't an issue at all, but if
you're using co-pilot which is also a client in this situation it will
fire metals requests to it which you don't want. We then throw an error
in that situation while the command still worked. This is confusing for
users, so this should fix the issue by filtering out the responses that
don't match metals.


ckipp01:
fixes #664

PR #702, opened 2024-12-14
fzf integration as an alternative to telescope - closes #701

### LINKED ISSUE #701: fzf-lua extension
### Describe the feature

LazyVim has now [replaced telescope with fzf-lua](https://www.lazyvim.org/news#added-plugins) by default.

The metals telescope extension is very, very handy, but is no longer compatible with LazyVim. It would be fantastic to have a [fzf-lua](https://github.com/ibhagwan/fzf-lua/wiki/Advanced#fzf-exec-api) flavour of it.

I'm giving it a try. No promise of results.

ornicar:
It seems to work. This is the most lua I have ever written, so it's probably all sorts of wrong.

For https://github.com/scalameta/nvim-metals/issues/701

image

### DIFF
--- a/doc/metals.txt
+++ b/doc/metals.txt
@@ -865,11 +865,11 @@ commands()                                                           *commands()
                            utilizing |vim.ui.select()|. If you're a telescope
                            user you can choose to use this with something like
                            https://github.com/stevearc/dressing.nvim or feel
-                           free to use the built-in telescope extension
-                           |metals-telescope| which just displays things a bit
-                           better since we have more control over using
-                           |vim.ui.select()|. If you're not a telescope user,
-                           then you'll want to use this.
+                           free to use the built-in extensions
+                           |metals-telescope| or |metals-fzf-lua| which just displays 
+                           things a bit better since we have more control over using
+                           |vim.ui.select()|. If you're not a telescope or
+                           fzf-lua user, then you'll want to use this.
 
                                                                *compile_cancel()*
 compile_cancel()           Use to execute a |metals.compile-cancel| command.
@@ -1370,7 +1370,7 @@ https://github.com/nvim-telescope/telescope.nvim which allows you to easily
 choose any of the |metals-commands|. You can trigger the picker by using the
 following function in a mapping: >
 
-  lua require("telescope").extensions.metals.commands()
+  :lua require("telescope").extensions.metals.commands()
 <
 
 If this is the only way you'll trigger the picker, then there is no need to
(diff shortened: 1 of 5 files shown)


ckipp01:
lThanks for sending this in @ornicar! However, I wonder if it's even needed. A while back I added the `commands()` function that you can use with 
```lua
require("metals").commands()
```
which will actually give you all of the commands metals knows about (minus the ones that are nvim-metals specific). A quick glance at the `fzf-lua` docs show me `register_ui_select` which will make `vim.ui.select` use `fzf-lua` and the `commands()` function is handled by whatever you have registered for `vim.ui.select`. What you have probably is a bit superior since it includes the extra couple commands for `nvim-metals`, but I still wanted to point it out. I don't have an issue merging this in this, especially if it's useful for those extra commands.

doc/metals.txt
```diff
@@ -1387,8 +1387,17 @@ longer be lazy loaded. >
...
-NOTE: If you'd rather utilize |vim.ui.select| for this a more generate
+Metals provides a custom picker from
+https://github.com/ibhagwan/fzf-lua which allows you to easily
+choose any of the |metals-commands|. You can trigger the picker by using the
+following function in a mapping: >
+
+  :lua require("metals").fzf_menu()
+<
+
+NOTE: If you'd rather utilize |vim.ui.select| for this, a more generic
```
ckipp01, line 1400: Ahhh I see what I wrote before you probably know since you touched this.

ornicar:
You're right! I overlooked that. I'll close the PR then, let's not add unnecessary code.

It was a good learning experience for me nonetheless :+1:

lua/fzf-lua/_extensions/metals.lua
```diff
@@ -0,0 +1,45 @@
...
+  for _, entry in ipairs(commands_table) do
+    if entry.label == label then
+      run_entry(entry)
+      return
+    end
+  end
+end
+
+return function(opts)
+  opts = opts or {}
```
ckipp01, line 29: I don't see one looking through the docs, but I wish there was a way to also include the description that would be shown but not considered when you search. Either way, maybe an improvement for later on if someone knows how to add that.

ckipp01 (approved):
Thanks @ornicar !

lua/fzf-lua/_extensions/metals.lua
ornicar: The only way I see is through a preview, as the fzf-lua menu entries are simple strings.

PR #711, opened 2025-01-30
chore: update nvim test versions, stylua, and selene

ckipp01:


ckipp01:
https://github.com/rhysd/action-setup-vim/pull/40 should land first

PR #713, opened 2025-02-01
feat: start using inlay_hints for worksheets

ckipp01:
This is a bit of an experiment to see if everything will fully work as expected in worksheet evaluations now that Metals is returning them as `inlayHint`s. There is a refresh bug it seems that should probably be addressed first because currently it's odd behavior where it ends up being stale, but then if you trigger the worksheet hover that shows the correct result even though the `inlayHint` itself is wrong. A couple things I'd like to do before merging this:

- [x] Get https://github.com/neovim/neovim/issues/32248 addressed if possible
- [x] Get https://github.com/neovim/neovim/issues/32283 addressed if possible
- [ ] Make some tests
- [x] ~Verify that we shouldn't be using resolve here. I created a message about this on Discord since we are always just returning the same stuff that we already have in the `inlayHint` so I'm making an assumption it's safe to just do it this way, but I really want to make sure this evaluation isn't ever different after a resolve.~ So I got some illustrations with newer Scala versions that do indeed use this correctly.


ofseed:
https://github.com/neovim/neovim/pull/32446 should resolve these problems.

ckipp01:
> [neovim/neovim#32446](https://github.com/neovim/neovim/pull/32446) should resolve these problems.

Awesome, thanks @ofseed! I just tested this and indeed it fixes both of them.

ofseed:
I'm currently exploring kinds of stuff related to inlay hints in neovim. The two bugs have existed for quite a while because they are heavily associated with `workspace/inlayHint/refresh` which other LSPs do not depend on except metals. So feel free to ping me if there are other bugs!

PR #716, opened 2025-02-04
Update hover_worksheet to use inlay hints

duartepinto:
This change was introduced in metals 1.5.0, which started using inlay hints instead of the previous custom solution.

### DIFF
--- a/lua/metals/decoration.lua
+++ b/lua/metals/decoration.lua
@@ -31,36 +31,31 @@ end
 
 M.hover_worksheet = function(opts)
   local buf = api.nvim_get_current_buf()
-  local line, _ = unpack(api.nvim_win_get_cursor(0))
+  local line,_ = unpack(api.nvim_win_get_cursor(0))
 
-  local exists, ext_ids = pcall(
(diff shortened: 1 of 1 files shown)


ckipp01:
Thanks for creating this @duartepinto! However there are a couple things I'd like to address before actually pulling the trigger on this. I actually already have a PR for this [here](https://github.com/scalameta/nvim-metals/pull/713) that you can try out. The main things are the two issues that I have linked upstream.

Have you noticed those while using this?

duartepinto:
> Have you noticed those while using this?

Weren't both of these issues already a problem, even before 1.5.0?

At least for _[LSP: textDocument/inlay_hint not being correctly published when received](https://github.com/neovim/neovim/issues/32248)_, whenever I used worksheets I felt the need to have to force the buffer to refresh so that the hints would be updated.

The second one ([LSP: textDocument/inlayHint at the end of a line not being removed when the line is deleted](https://github.com/neovim/neovim/issues/32283)), I can't say I've seen it using worksheets, but I've seen a similar behavior happening when deleting lines that are being decorated with errors or warnings.

PR #717, opened 2025-02-19
Fix execute_command callback mechanism

aPisC:
The `execute_command` function in `lua/metals.lua` has an issue, where it ignores the response from metals in some cases and never invokes the callback even if the response returned successfully. This occurres in those cases where metals is not the only LSP client attached to the buffer (Other one was copilot in my case), and the other client gets a smaller `client_id` than metals. The `buf_request_all` notifies all clients with the requested action and the other client's response comes before metals' so the for loop exits the whole response handling because the `client_id` mismatch. 

Possible solution is to check explicitly if metals has a response (with its `client_id`) and use it to invoke the callback.


ckipp01 (approved):
This looks good, thanks for catching this @aPisC!

PR #727, opened 2025-03-31
Fix worksheet hover output not being in markdown

EarlPitts:
Hello!

I recently started using the new inlay hints feature for worksheets, and I noticed that the hover window's output looks a bit strange:
image

I noticed that even though the hover window's syntax is set to markdown, the result itself is not wrapped in backticks, so it's just displayed as plain text.
Wrapping it makes the result display in the same way as it used to:
image

I'm not that familiar with lua, so if this is not the most idiomatic way to do this, I'm open for suggestions. 😄

### DIFF
--- a/lua/metals.lua
+++ b/lua/metals.lua
@@ -493,8 +493,10 @@ M.hover_worksheet = function(opts)
   if evaluation_result then
     local tooltip = evaluation_result.inlay_hint.tooltip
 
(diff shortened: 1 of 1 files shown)


EarlPitts:
I also noticed that in some cases there is no value in the tooltip (e.g.: `IO.unit.unsafeRunSync()`), that results in an error, so I also added a check for this.

ckipp01 (approved):
Thanks @EarlPitts !

PR #738, opened 2025-06-03
chore: support new Metals config options

susliko:
Hi 👋🏻 
Encouraged by [Metals v1.5.3](https://scalameta.org/metals/blog/2025/05/13/strontium#best-effort-compilation) release notes I wanted to try best-effort compilation and noticed that `enableBestEffort` option and a few others were missing from the list of valid options.

As a reference I used [UserConfiguration](https://github.com/scalameta/metals/blob/e4f70f9ca82ca96f4e72013b68cd3d7f82c07b79/metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala#L29) case class, but I'm not entirely sure if it is contains a comprehensive list of options since options like `ammoniteJvmProperties`, `javaFormat.eclipseConfigPath`, `showImplicitArguments` are not there 🤔


ckipp01:
Thanks @susliko, yes we are missing some here. Let me look at these in Metals to ensure there is nothing else that might be needed with this. We'll also want to try and add these to the docs here as well. I can do this on top of your PR as well.

PR #767, opened 2026-01-13
Add more required VM options for Metals v2

olafurpg:
Previously, nvim-metals would crash if it ran the new 2.0.0-M3 milestone
release because it was missing VM options.
```
2026.01.12 16:17:39 INFO  Started: Metals version 2.0.0-M2-3-fb622b19-SNAPSHOT in folders '/Users/username/work/bazelbuild/bazel' for client Neovim 0.12.0-dev+g8d5452c46d.[0m
Exception in thread "Compiler Job Thread [javac-default-java]" java.lang.IllegalAccessError: class scala.meta.internal.jpc.JavaPruneCompiler (in unnamed module @0x1187c9e8) cannot access class com.sun.tools.javac.api.JavacTool (in module jdk.compiler) because module jdk.compiler does not export com.sun.tools.javac.api to unnamed module @0x1187c9e8
	at scala.meta.internal.jpc.JavaPruneCompiler.<init>(JavaPruneCompiler.scala:57)
```
Now, nvim-metals will include these VM options out of the box so that it
no longer crashes.

## Implementation Note

In VS Code, the extension actually reads the
META-INF/metals-required-vm-options.txt resource file from the published
Metals so that it picks up new required VM options in future releases
(see https://github.com/scalameta/metals-vscode/pull/1818). I
initially tried to implement the same approach here but didn't like the
solution, so this PR implements the simpler approach of just
hardcoding the list. This list should be pretty stable across releases
so it probably won't need to update frequently.

- → **#767** ← [[Commit](https://github.com/scalameta/nvim-metals/pull/767/commits/6ccdf3d8b06c20f83f4506e1e5e55f7d75966c99)]

### DIFF
--- a/lua/metals/config.lua
+++ b/lua/metals/config.lua
@@ -388,7 +388,42 @@ local function validate_config(config, bufnr)
 
   local passed_in_options = config.settings.metals.serverProperties or {}
 
-  local all_opts = util.merge_lists(passed_in_options, valid_java_opts)
+  -- Required VM options for Metals v2.x (from resource file META-INF/metals-required-vm-options.txt).
+  -- These are needed for JDK internals access (javac APIs, JOL, etc.).
+  -- Added first so they can be overridden by user settings if needed.
+  --
+  -- To update these options for a new Metals version, run:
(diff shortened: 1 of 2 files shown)


olafurpg:
Let me remove the accidental diffs :melting_face:

Issue #768, opened 2026-01-15
`MetalsUpdate` installs unstable server version


usommerl:
I occasionally run `MetalsUpdate` to make sure I have an up-to-date Metals installation. [According to the documentation][0], this command should install the latest stable release unless a custom version is specified via the `serverVersion` setting.

When I ran the command today, I had not set `serverVersion`. The nvim-metals plugin was checked out at commit [90dadd6][1], and the latest stable Metals release at the time was 1.6.4. Despite this, `MetalsUpdate` installed the milestone version [2.0.0-M6][3] instead. This completely broke my development setup, yet it took me some time to realize what was wrong because the Metals integration still appeared to be partially working.

I’m not very familiar with the nvim-metals codebase, but perhaps this issue could be avoided by assigning `latest_stable = "latest.stable"` instead of `latest_stable = "latest.release"` [here][4]?

<img width="1278" height="287" alt="Image" src="https://github.com/user-attachments/assets/d9b82579-ca80-4f75-ac9c-301e44d40ccf" />

[0]: https://github.com/scalameta/nvim-metals/blob/90dadd67bb1069ce1ba2a4dabf4e45da2807ba80/doc/metals.txt#L826-L829
[1]: https://github.com/scalameta/nvim-metals/tree/90dadd67bb1069ce1ba2a4dabf4e45da2807ba80
[2]: https://github.com/scalameta/metals/releases/tag/v1.6.4
[3]: https://github.com/scalameta/metals/releases/tag/v2.0.0-M6
[4]: https://github.com/scalameta/nvim-metals/blob/90dadd67bb1069ce1ba2a4dabf4e45da2807ba80/lua/metals/install.lua#L8

ckipp01:
Thanks for the report @usommerl, yes I didn't realize this and it needs to be addressed. There is ongoing work of a 2.x series of Metals that is quite different than the 1.x series, and yes, we should stick with one here. However I need to figure out out to handle this because even using `latest.stable` might bight us if 2.x is released before we want to actually switch to that.

PR #769, opened 2026-01-16
fix: use latest.stable to avoid pulling v2 for now

ckipp01:


ckipp01:
Fixes https://github.com/scalameta/nvim-metals/issues/768 for now.

PR #781, opened 2026-04-26
I have added root dir to MetalsInfo and made some changes to root_patterns

torje:
MetalsInfo: I have added what I think is useful info: root_dir.

Dropped: With regard to root_patterns: I believe it is an improvement, but I have no experience on other build systems than gradle, so I could easily be wrong on the root_pattern removal.

### DIFF
--- a/lua/metals.lua
+++ b/lua/metals.lua
@@ -167,7 +167,13 @@ M.info = function()
   else
     table.insert(output, string.format("%s %s", loc_msg, conf.metals_bin()))
   end
+  if config then
+    table.insert(output, string.format("  - metals root dir: %s", config.root_dir))
+  else
+    table.insert(output, "  - metals config is nil")
+  end
   table.insert(output, "")
+
   table.insert(output, "## Helpful links")
   table.insert(output, "  - https://discord.gg/FaVDrJegEh")
   table.insert(output, "  - https://matrix.to/#/#scalameta:vim-users")


lua/metals/config.lua
```diff
@@ -335,8 +335,8 @@ local function validate_config(config, bufnr)
       "build.sc",
       "build.mill",
       "settings.gradle",
+      "gradlew",
```
ckipp01: I'm not a huge gradle user, but it was my understanding that every gradle project will have a `settings.gradle`, or a `settings.gradle.kts` at the root. Shouldn't that be enough here since we already have it in the patterns? Or do you have a usecase where you don't have a `settings.gradle`, but do have a `gradlew`?

lua/metals/config.lua
```diff
@@ -335,8 +335,8 @@ local function validate_config(config, bufnr)
...
       "pom.xml",
-      ".scala-build",
```
ckipp01, line 339: How come this was removed? For scala-cli projects this is still useful since it will be generated at the root of your project.

ckipp01 (changes requested):
Thanks for the PR! Just a couple question though.

lua/metals/config.lua
torje: Then that is a misunderstanding from my side. I got one of those deeper withing the structure, but probably me doing something wrong then. Will remove that change.

lua/metals/config.lua
torje: I should have tested this before, yes. We can drop this whole commit.

lua/metals/config.lua
ckipp01: Yea, this one is a bit tricky as you can in theory have a project and then a subdirectory that has scripts in it which is detected here. There isn't a great way around this apart from doing a multi-root project in Metals.

ckipp01 (approved):
Thanks!

PR #783, opened 2026-06-16
Update `toggle_setting` to handle nested booleans like inlayHints

henryxparker:
Reason:
Metals supports dynamically changing the `inlayHints` settings, but the current `toggle_setting` did not work with the nested structure.
Specifically, it did not have any of the inlayHints in the "accepted" settings, and even if they were to be added, it still would fail because `settings["inlayHints.implicitArguments.enable"] =/= settings.inlayHints.implicitArguments.enable`.
Also I think the check for if the setting was a boolean was doing nothing, because it was checking the string itself; so it was doing `not type(string) == "boolean"` -> `not 'string' == boolean` -> `false == "boolean` -> `false`(always). The code can't know if the setting is a boolean unless it was already set, but neither could the old code so ¯\\\_(ツ)_/¯.

I changed the code so it interprets any `.` as a nested table. There are already accepted settings that have a `.` like `java-format.eclipse-profile`, but both of the `java-format` options are not boolean, and they might even be nested tables like this code expects.


ckipp01 (approved):
Thanks!

